PR #93, opened 2026-07-09
fix: guard GO_LIVE and CUT_STREAM against unactivated productions (closes #87)

### LINKED ISSUE #87: [SECURITY][LOW] GO_LIVE/CUT_STREAM WebSocket messages update CouchDB with no production flow validation
## Summary
The `GO_LIVE` and `CUT_STREAM` WebSocket message handlers update `doc.status` and broadcast `ON_AIR` events without verifying that a Strom flow is actually running. Any client holding a valid WebSocket connection can flip the on-air status — there is no production-level RBAC beyond the shared API key. Setting `status: 'active'` via `GO_LIVE` when no Strom flow is running creates a permanent inconsistency where `status=active` but `stromFlowId` is undefined, which can break subsequent activate/deactivate logic.

## CVSS Score
4.3 (Low) — CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:N/I:L/A:L

## Category
OWASP A01 — Broken Access Control

## Affected File(s)
- `src/ws/controller.ts:751-762`

## Steps to Reproduce
1. Connect to `ws://host/api/v1/productions/:id/controller?key=API_KEY` without activating the production
2. Send `{ "type": "GO_LIVE" }`
3. Production status changes to `active` even though no pipeline is running

## Remediation
At minimum, validate that a Strom flow exists before writing:
```typescript
case 'GO_LIVE': {
  if (!doc.stromFlowId) {
    ws.send(JSON.stringify({ type: 'ERROR', error: 'Production is not activated' }));
    break;
  }
  // proceed
}
```
For full protection, consider a distinct operator role with RBAC on these state-changing messages.

## References
- [OWASP A01:2021 — Broken Access Control](https://owasp.org/Top10/A01_2021-Broken_Access_Control/)

LucasMaupin:
## Summary

- `src/ws/controller.ts`: both `GO_LIVE` and `CUT_STREAM` handlers now check `doc.stromFlowId` before writing to CouchDB and broadcasting tally events
- If the production is not activated (no Strom flow running), an `ERROR` message is sent back to the WebSocket client and the state is not changed

## Why

Without this guard, any WebSocket client with a valid API key could send `GO_LIVE` when no pipeline was running, setting `doc.status = 'active'` while `doc.stromFlowId` remained undefined. This left the production in a permanently inconsistent state that could break subsequent `activate`/`deactivate` operations. The same logic flaw applied to `CUT_STREAM`.

Closes #87

### DIFF
--- a/src/ws/controller.ts
+++ b/src/ws/controller.ts
@@ -749,12 +749,20 @@ async function handleMessage(
       break;
     }
     case 'GO_LIVE': {
+      if (!doc.stromFlowId) {
+        ws.send(JSON.stringify({ type: 'ERROR', error: 'Production is not activated' }));
+        break;
+      }
       const updated: ProductionDoc = { ...doc, status: 'active', updatedAt: new Date().toISOString() };
       await db.insert(updated);
       broadcast(productionId, { type: 'ON_AIR', value: true });
       break;
     }
     case 'CUT_STREAM': {
+      if (!doc.stromFlowId) {
+        ws.send(JSON.stringify({ type: 'ERROR', error: 'Production is not activated' }));
+        break;
+      }
       const updated: ProductionDoc = { ...doc, status: 'active', updatedAt: new Date().toISOString() };
       await db.insert(updated);
       broadcast(productionId, { type: 'ON_AIR', value: false });


birme (commented):
APPROVE — doc.stromFlowId is a real optional field (db/types.ts:142); guarding GO_LIVE/CUT_STREAM on it correctly prevents flipping status to active with no running flow, addressing #87. Note (pre-existing, out of scope): CUT_STREAM still sets status:'active' rather than an off-air state — the PR faithfully preserves prior behavior; consider a follow-up. Co-Authored-By trailer present. No fabrication.

PR #95, opened 2026-07-10
fix: rate-limit /reconnect to 5/min and remove cleanLegacyFixtures from reconnect handler (closes #89)

### LINKED ISSUE #89: [SECURITY][LOW] POST /api/v1/reconnect has no rate limit and executes cleanLegacyFixtures on every call — DoS vector
## Summary
The `/api/v1/reconnect` endpoint is auth-exempt and has no specific rate limit beyond the global 200/min. On each call it attempts `connectDb()` (opening a new Nano connection and listing all CouchDB databases) and calls `cleanLegacyFixtures()` (up to five CouchDB reads and destroys any matching documents). An unauthenticated attacker can call this endpoint 200 times per minute, each triggering multiple CouchDB operations and a Strom version check, causing database and Strom CPU exhaustion. This is distinct from issue #59 (which tracks that the endpoint leaks infrastructure status); this issue is about the rate limiting and side-effect of running cleanLegacyFixtures on every call.

## CVSS Score
3.7 (Low) — CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:L

## Category
OWASP A04 — Insecure Design

## Affected File(s)
- `src/routes/status.ts:22-46`

## Steps to Reproduce
```bash
for i in $(seq 200); do curl -X POST http://api/api/v1/reconnect & done
# Triggers 200+ concurrent CouchDB connection attempts and cleanLegacyFixtures executions
```

## Remediation
1. Apply a tight per-endpoint rate limit (3–5 requests/minute)
2. Move the legacy-fixture cleanup to startup only — not every reconnect call:
```typescript
// In startup.ts:
await cleanLegacyFixtures();

// In status.ts reconnect handler — remove the cleanLegacyFixtures() call
await connectDb();
// No legacy fixture cleanup here
```

## References
- [OWASP A04:2021 — Insecure Design](https://owasp.org/Top10/A04_2021-Insecure_Design/)
- [CWE-400: Uncontrolled Resource Consumption](https://cwe.mitre.org/data/definitions/400.html)

LucasMaupin:
## Summary

- `src/routes/status.ts`: adds a per-route rate limit of 5 requests/minute to `POST /api/v1/reconnect` via `@fastify/rate-limit`'s route-level config
- `src/routes/status.ts`: removes the `cleanLegacyFixtures()` call from the reconnect handler

## Why

The `/reconnect` endpoint is auth-exempt and was only limited by the global 200 req/min cap. Each call triggered `connectDb()` (a new Nano connection + CouchDB database list) and `cleanLegacyFixtures()` (up to five CouchDB reads and potential document destroys), making it a cheap DoS vector.

`cleanLegacyFixtures()` already runs at startup in `main.ts` — there is no need to repeat it on every reconnect. The reconnect path now only re-establishes the DB connection when disconnected, as intended.

The per-route limit of 5/min is appropriate for a recovery endpoint that should rarely be called and never in rapid bursts.

Closes #89

### DIFF
--- a/src/routes/status.ts
+++ b/src/routes/status.ts
@@ -3,7 +3,6 @@ import { isDbReady, connectDb, isDbConnected } from '../db/index.js';
 import { StromClient } from '../lib/strom.js';
 import { getStromToken } from '../lib/strom-token.js';
 import { config } from '../config.js';
-import { cleanLegacyFixtures } from '../db/seed.js';
 
 const statusRoutes: FastifyPluginAsync = async (fastify) => {
   fastify.get('/api/v1/ping', async (_req, reply) => {
@@ -19,13 +18,15 @@ const statusRoutes: FastifyPluginAsync = async (fastify) => {
     }
   });
 
-  fastify.post('/api/v1/reconnect', async (_req, reply) => {
+  fastify.post(
+    '/api/v1/reconnect',
+    { config: { rateLimit: { max: 5, timeWindow: '1 minute' } } },
+    async (_req, reply) => {
     const result: { db: boolean; strom: boolean } = { db: isDbConnected(), strom: false };
 
     if (!result.db) {
       try {
         await connectDb();
-        await cleanLegacyFixtures();
         fastify.log.info('[reconnect] Database connection restored');
         result.db = true;
       } catch (err: any) {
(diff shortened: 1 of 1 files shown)


birme (commented):
APPROVE — Route-level `config: { rateLimit: { max: 5, timeWindow: '1 minute' } }` is valid @fastify/rate-limit usage (plugin registered globally in src/server.ts). Removing cleanLegacyFixtures() from the reconnect handler is safe: it still runs at startup in src/main.ts:92. Directly addresses the unauthenticated DoS in issue #89. Commit has Co-Authored-By trailer.

PR #97, opened 2026-07-11
fix: coalesce concurrent SAT token exchange requests to prevent burst PAT exhaustion (closes #85)

### LINKED ISSUE #85: [SECURITY][MEDIUM] SAT token cache concurrent-fetch race — parallel requests each independently exchange the PAT
## Summary
`getStromToken` checks the module-level cache and, on a miss, fires an independent token exchange for each concurrent caller. If the SAT just expires and 10 WebSocket messages arrive simultaneously, all 10 callers independently POST to `token.svc.prod.osaas.io`. Beyond redundant network calls, if the OSC token service enforces per-PAT rate limits, burst activation across multiple productions could exhaust those limits and cause all subsequent Strom calls to fail with 401.

## CVSS Score
5.3 (Medium) — CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:N/I:N/A:H

## Category
OWASP A04 — Insecure Design / Race Condition

## Affected File(s)
- `src/lib/strom-token.ts:42-63`

## Steps to Reproduce
Trigger parallel activation of 10 productions simultaneously after the SAT expires — observe N identical token exchange calls in network logs.

## Remediation
Implement promise coalescing to ensure only one in-flight exchange at a time:
```typescript
let inflightExchange: Promise<string> | null = null;

export async function getStromToken(): Promise<string> {
  if (cache && !isExpiringSoon(cache)) return cache.token;
  if (inflightExchange) return inflightExchange;
  inflightExchange = fetchNewToken()
    .then((token) => { cache = { token, expiresAt: ... }; return token; })
    .finally(() => { inflightExchange = null; });
  return inflightExchange;
}
```

## References
- [OWASP Race Condition Prevention](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html)
- [CWE-362: Race Condition](https://cwe.mitre.org/data/definitions/362.html)

LucasMaupin:
## Summary

- Adds an `inflightExchange` promise guard to `getStromToken` in `src/lib/strom-token.ts`
- Concurrent callers that arrive after SAT expiry now await the single shared in-flight promise instead of each independently POSTing to the OSC token service
- Once the exchange completes the shared promise is cleared so future expirations trigger a fresh exchange

## Why

When multiple productions activate simultaneously right after a SAT expires, each parallel call to `getStromToken` independently fires a `POST` to `token.svc.prod.osaas.io`. If OSC enforces per-PAT rate limits, the burst can exhaust them and cause all subsequent Strom calls to fail with 401. Promise coalescing ensures only one exchange is ever in flight at a time.

## Test

No automated tests exist for this module yet. Manual verification: trigger parallel `getStromToken` calls after cache expiry and confirm only one network request is made.

Closes #85

### DIFF
--- a/src/lib/strom-token.ts
+++ b/src/lib/strom-token.ts
@@ -17,6 +17,8 @@ interface SatCache {
 }
 
 let cache: SatCache | null = null
+// Coalesce concurrent token exchange requests to avoid burst PAT exhaustion
+let inflightExchange: Promise<string> | null = null
 
 function isExpiringSoon(cache: SatCache): boolean {
   return Date.now() >= cache.expiresAt - REFRESH_BUFFER_MS
@@ -43,7 +45,14 @@ export async function getStromToken(pat: string | undefined): Promise<string | u
     return cache.token
   }
 
-  const res = await fetch(TOKEN_EXCHANGE_URL, {
+  // Coalesce: if an exchange is already in-flight, wait for it instead of
+  // issuing a duplicate request. This prevents burst activations from
+  // exhausting OSC per-PAT rate limits after a SAT expiry.
+  if (inflightExchange) {
+    return inflightExchange
+  }
+
+  inflightExchange = fetch(TOKEN_EXCHANGE_URL, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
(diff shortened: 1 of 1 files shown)


birme (commented):
APPROVE — Promise-coalescing via inflightExchange correctly serializes token exchange, addressing the concurrent-fetch race in #85. Token response shape {token, expiry} and cache {token, expiresAt=expiry*1000} match the existing code exactly (strom-token.ts:61-62); error handling (res.ok check) is preserved inside the .then, and .finally clears the guard so a failed exchange doesn't wedge future calls. No fabricated API. Minor: commit is missing a Co-Authored-By trailer per repo convention — please add it. Approving on correctness.

PR #104, opened 2026-07-15
fix(security): API_KEY gate for /ws/ production controller

### LINKED ISSUE #101: [SECURITY][HIGH] WebSocket /ws/ path bypasses API key auth guard — startsWith('/api/v1') check skips all WS connections
## Summary

The API key authentication hook in `server.ts` only guards routes whose URL starts with `/api/v1`. The WebSocket controller is registered at `/ws/productions/:id/controller`. Because the guard condition checks `!req.url.startsWith('/api/v1')` and returns early if false, all `/ws/` paths silently bypass the authentication hook entirely — even when `API_KEY` is configured. Any unauthenticated client can connect and send CUT, TAKE, AUDIO_SET, DSK_TOGGLE, and other live production commands.

## CVSS Score
7.5 (High) — CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:H/A:L

## Category
OWASP A01:2021 – Broken Access Control

## Affected File(s)
- `src/server.ts` — auth `onRequest` hook (url prefix guard)
- `src/ws/controller.ts` — WebSocket handler registered at `/ws/productions/:id/controller`

## Steps to Reproduce
1. Deploy Open Live with `API_KEY=secret` configured.
2. Connect a WebSocket client to `ws://<host>/ws/productions/<id>/controller` with no `Authorization` header and no `?key=` query parameter.
3. Send a `CUT` or `TAKE` JSON message.
4. Observe that the command is accepted and executed without any auth challenge.

## Remediation
Extend the auth hook to cover the `/ws/` prefix as well:

```ts
// server.ts — in the onRequest auth hook
if (!req.url.startsWith('/api/v1') && !req.url.startsWith('/ws/')) return;
```

Or, preferably, move WebSocket routes under `/api/v1/ws/` so they inherit the existing guard automatically. Additionally, validate the API key inside the WebSocket upgrade handler as a defence-in-depth measure.

## References
- Supersedes closed #7 (fix was incomplete — the `/api/v1` prefix guard does not cover `/ws/` paths)
- OWASP A01:2021 – https://owasp.org/Top10/A01_2021-Broken_Access_Control/

YRWoods:
## Summary
HIGH: WebSocket controller under `/ws/` bypassed the API key hook (only `/api/v1` was checked).

## Approach
- Extend `onRequest` auth to `/ws/`
- Defence-in-depth: reject unauthorized sockets in the WS upgrade handler
- Document `?key=` / Bearer for WS

## Fixes
#101

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -27,8 +27,9 @@ export const config = {
   stromAuthMode: (process.env['STROM_AUTH_MODE'] ?? 'osc') as 'osc' | 'direct',
   logLevel: process.env['LOG_LEVEL'] ?? 'info',
   /**
-   * Optional static API key. When set, all /api/v1 routes require:
+   * Optional static API key. When set, /api/v1 and /ws/ routes require:
    *   Authorization: Bearer <API_KEY>
+   * WebSocket upgrades may also pass ?key=<API_KEY>.
    * Leave unset when running behind OSC's reverse-proxy auth wall.
    */
   apiKey: process.env['API_KEY'] ?? undefined,
--- a/src/server.ts
+++ b/src/server.ts
@@ -107,12 +107,14 @@ export async function buildServer() {
 
   // Optional API key authentication — enabled when API_KEY env var is set.
   // Exempt: health/ready probes and status/reconnect endpoints.
-  // WS connections: pass key via Authorization header or ?key= query param on upgrade.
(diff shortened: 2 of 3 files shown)


YRWoods:
Withdrawing.

Issue #88, opened 2026-07-06
[SECURITY][MEDIUM] Production values record forwards arbitrary string keys to Strom block properties — property injection


LucasMaupin:
## Summary
`ProductionPatch` accepts `values` as `z.record(z.union([z.string(), z.number(), z.boolean()]))` — any key/value combination is allowed. Several values are forwarded directly to Strom block properties: `pgm_resolution`, `pgm_framerate`, `clock` (forwarded verbatim as `clock_type`), `num_pips`, and more. None of these string values are validated against an allowlist of known-safe patterns. An attacker can inject malformed strings into flow properties sent to Strom, potentially exploiting parsing vulnerabilities in Strom's backend or triggering unexpected GStreamer pipeline behaviour. This is distinct from issue #60 (which covers `pgm_resolution` / `pgm_framerate` forwarding raw); this issue covers the entire unvalidated `values` record including `clock` type and other keys.

## CVSS Score
6.1 (Medium) — CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:C/C:N/I:L/A:L

## Category
OWASP A03 — Injection

## Affected File(s)
- `src/routes/productions.ts:317`
- `src/lib/flow-generator.ts:118-163`

## Steps to Reproduce
```bash
curl -X PATCH http://api/api/v1/productions/:id \
  -H "Authorization: Bearer KEY" \
  -H "Content-Type: application/json" \
  -d '{"values": {"clock": "ntp;\x00DROP TABLE flows;--"}}'
# Then activate the production — the malformed clock_type is sent to Strom
```

## Remediation
Validate string values against format-specific allowlists before use:
```typescript
const RESOLUTION_RE = /^\d{3,5}x\d{3,5}$/;
const FRAMERATE_RE = /^\d{1,3}\/\d{1,3}$|^\d{1,3}$/;
const CLOCK_TYPES = new Set(['ntp', 'gst', 'system']);

// In flow-generator.ts before forwarding to Strom:
if (values.clock && !CLOCK_TYPES.has(String(values.clock))) {
  throw new Error(`Invalid clock type: ${values.clock}`);
}
```

## References
- [OWASP A03:2021 — Injection](https://owasp.org/Top10/A03_2021-Injection/)
- [CWE-20: Improper Input Validation](https://cwe.mitre.org/data/definitions/20.html)

LucasMaupin:
[...]

Issue #86, opened 2026-07-06
[SECURITY][MEDIUM] GRAPHIC_ON/GRAPHIC_OFF accept arbitrary overlayId — no verification against stored graphic IDs


LucasMaupin:
## Summary
The `GRAPHIC_ON` and `GRAPHIC_OFF` WebSocket message handlers search `doc.graphics` for a matching overlay ID and persist the updated document without verifying that the supplied `overlayId` references a real overlay in the production's graphic inventory. Sending a nonexistent ID still triggers a CouchDB write (a no-op tally mutation), generating spurious audit log entries and unnecessary database writes. Additionally, concurrent writes can produce CouchDB 409 conflicts that propagate as unhandled exceptions.

## CVSS Score
5.3 (Medium) — CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:N/I:L/A:L

## Category
OWASP A03 — Injection / A01 — Broken Access Control

## Affected File(s)
- `src/ws/controller.ts:763-775`

## Steps to Reproduce
1. Connect to the production WebSocket with a valid API key
2. Send `{ "type": "GRAPHIC_ON", "overlayId": "nonexistent-id-xyz" }`
3. Observe the server writes an updated CouchDB document even though no graphic matched

## Remediation
Verify the overlay ID exists before writing:
```typescript
case 'GRAPHIC_ON': {
  const graphic = doc.graphics.find((g) => g.id === msg.overlayId);
  if (!graphic) {
    ws.send(JSON.stringify({ type: 'ERROR', error: 'Overlay not found' }));
    break;
  }
  // proceed with update
}
```

## References
- [OWASP A01:2021 — Broken Access Control](https://owasp.org/Top10/A01_2021-Broken_Access_Control/)
- [CWE-20: Improper Input Validation](https://cwe.mitre.org/data/definitions/20.html)

LucasMaupin:
GRAPHIC_ON/OFF write to CouchDB even for non-existent overlay IDs, generating spurious writes and risking 409 conflicts. Labelled `bug` and `good-first-issue` — adding an existence check before the update in `controller.ts` is a well-scoped change. A contributor can pick this up.

LucasMaupin:
PR opened: https://github.com/Eyevinn/open-live/pull/98

PR #105, opened 2026-07-15
fix(security): validate production values + graphic overlay IDs

YRWoods:
## Frontier security

- **#88** — allowlist/format-check `production.values` (clock, resolution, framerate, …) before Strom
- **#86** — GRAPHIC_ON/OFF require real overlay id; no no-op CouchDB writes

Also includes prior WS API_KEY gate on this branch lineage if needed separately.


YRWoods:
Withdrawing.

PR #107, opened 2026-07-15
fix(security): stop leaking internal errors on activate/deactivate

### LINKED ISSUE #79: [SECURITY][MEDIUM] Activate/deactivate per-route catch blocks bypass global error handler — raw internal error messages returned in 500 responses
## Summary

The global Fastify error handler registered in `src/server.ts` correctly strips internal details from 5xx responses. However, two per-route `try/catch` blocks in `src/routes/productions.ts` — for the `POST /:id/activate` (line ~533) and `POST /:id/deactivate` (line ~587) endpoints — catch exceptions and forward `err.message` directly to the client:

```typescript
const message = err instanceof Error ? err.message : String(err)
return reply.status(500).send({ error: message, statusCode: 500 })
```

These per-route handlers run **before** the global error handler and short-circuit it. The raw `err.message` can contain Strom API error bodies (including internal service URLs), CouchDB error descriptions with document IDs, DNS resolution errors exposing internal hostnames, and SAT exchange failure details exposing the token service endpoint at `token.svc.prod.osaas.io`. This is a regression of the fix applied in #12 — the global handler was hardened but these two routes were not updated.

## CVSS Score
5.3 (Medium) — CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:N/A:N

## Category
OWASP A09:2021 — Security Logging and Monitoring Failures / Information Disclosure

## Affected File(s)
- `src/routes/productions.ts` — activate route error catch (~line 533–535)
- `src/routes/productions.ts` — deactivate route error catch (~line 587–589)

## Steps to Reproduce
1. Force a Strom API failure during activation (e.g. misconfigured `STROM_URL`).
2. `POST /api/v1/productions/:id/activate` — receive 500 with raw error message including internal service details.

## Remediation

Replace the raw `err.message` forwarding with a generic client message and log the full error server-side:

```typescript
// Before (leaks internal details):
const message = err instanceof Error ? err.message : String(err)
return reply.status(500).send({ error: message, statusCode: 500 })

// After:
req.log.error({ err }, 'Activation failed')
return reply.status(500).send({ error: 'Activation failed — check server logs', statusCode: 500 })
```

Apply the same pattern to the deactivate catch block. This aligns these routes with the global error handler pattern established in #12.

## References
- [OWASP A09:2021 — Security Logging and Monitoring Failures](https://owasp.org/Top10/A09_2021-Security_Logging_and_Monitoring_Failures/)
- Related: #12 (global error handler hardening)

YRWoods:
## Summary
Closes #79.

Per-route catch blocks on activate/deactivate forwarded `err.message` to clients, short-circuiting the global Fastify error handler and leaking Strom/CouchDB/SAT internals.

## Change
- Log full error server-side via `req.log.error`
- Return generic 500 messages only

## Test plan
- [ ] Force activation failure — response body has no internal host/URL/token detail
- [ ] Server logs still contain full error
- [ ] Deactivate path same

### DIFF
--- a/src/routes/productions.ts
+++ b/src/routes/productions.ts
@@ -530,9 +530,12 @@ const productionsRoutes: FastifyPluginAsync = async (fastify) => {
         _rev: insertResponse.rev,
       });
     } catch (err) {
-      const message = err instanceof Error ? err.message : String(err);
-      fastify.log.error({ err }, 'Failed to initiate production activation');
-      return reply.status(500).send({ error: message, statusCode: 500 });
+      // Do not forward err.message to clients — can leak Strom/CouchDB/SAT internals (#79)
+      req.log.error({ err }, 'Failed to initiate production activation');
+      return reply.status(500).send({
+        error: 'Activation failed — check server logs',
+        statusCode: 500,
+      });
     }
   });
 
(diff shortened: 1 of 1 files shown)


YRWoods:
Withdrawing.

PR #110, opened 2026-07-15
fix(security): validate production.values and gate controller WS

YRWoods:
## Summary
- Validate `production.values` (known keys, resolution/framerate/clock bounds) on PATCH and again at Strom flow build.
- Reject unknown graphic overlay IDs on GRAPHIC_ON/OFF and macros.
- When `API_KEY` is set, require Bearer or `?key=` on controller WebSocket upgrades (parity with HTTP).

## Why
Defence-in-depth for values that reach Strom block properties; close WS auth gap when static API key is used.

## Test plan
- [ ] `npm run typecheck`
- [ ] PATCH production with invalid `pgm_resolution` → 400
- [ ] GRAPHIC_ON with unknown overlayId → ERROR
- [ ] With `API_KEY` set, WS without key closes Unauthorized; with `?key=` connects

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -27,8 +27,9 @@ export const config = {
   stromAuthMode: (process.env['STROM_AUTH_MODE'] ?? 'osc') as 'osc' | 'direct',
   logLevel: process.env['LOG_LEVEL'] ?? 'info',
   /**
-   * Optional static API key. When set, all /api/v1 routes require:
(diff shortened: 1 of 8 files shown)


YRWoods:
Closing as overlapping earlier scoped PRs: #104 (WS API key) and #105 (production values + graphic IDs). Prefer those smaller reviews over this combined branch. New distinct ship: reconnect rate-limit for #89.

PR #111, opened 2026-07-15
fix(security): rate-limit reconnect and skip fixture cleanup

### LINKED ISSUE #89: [SECURITY][LOW] POST /api/v1/reconnect has no rate limit and executes cleanLegacyFixtures on every call — DoS vector

YRWoods:
## Summary
Closes #89.

- Cap `POST /api/v1/reconnect` at **5 req/min** (auth-exempt DoS surface).
- Stop calling `cleanLegacyFixtures()` on every reconnect (still runs at startup in `main.ts`).
- Reconnect still reconnects DB + probes Strom version.

## Why
Unauthenticated clients could hit global 200/min and force repeated CouchDB list/destroy work via fixture cleanup.

## Test plan
- [x] `npm run typecheck`
- [ ] `POST /api/v1/reconnect` when DB up → 200/503 as before
- [ ] 6th call within a minute → 429

### DIFF
--- a/src/routes/status.ts
+++ b/src/routes/status.ts
@@ -3,7 +3,6 @@ import { isDbReady, connectDb, isDbConnected } from '../db/index.js';
 import { StromClient } from '../lib/strom.js';
 import { getStromToken } from '../lib/strom-token.js';
 import { config } from '../config.js';
-import { cleanLegacyFixtures } from '../db/seed.js';
 
 const statusRoutes: FastifyPluginAsync = async (fastify) => {
   fastify.get('/api/v1/ping', async (_req, reply) => {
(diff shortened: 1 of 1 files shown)


YRWoods:
Withdrawing.

PR #122, opened 2026-07-15
fix(security): coalesce concurrent SAT token exchanges

### LINKED ISSUE #85: [SECURITY][MEDIUM] SAT token cache concurrent-fetch race — parallel requests each independently exchange the PAT

YRWoods:
## Summary
Closes #85.

Only one PAT→SAT exchange in flight at a time; concurrent callers await the same promise.

## Test plan
- [x] `npm run typecheck`

### DIFF
--- a/src/lib/strom-token.ts
+++ b/src/lib/strom-token.ts
@@ -5,6 +5,8 @@
  * - STROM_AUTH_MODE=osc (default): exchanges the PAT for a short-lived SAT
  *   via the OSC token service (required for OSC-hosted Strom instances).
  *   Caches the SAT and refreshes it 5 min before expiry.
+ *
+ * #85: coalesce concurrent exchanges so only one in-flight fetch runs at a time.
  */
 
 const TOKEN_EXCHANGE_URL = 'https://token.svc.prod.osaas.io/servicetoken'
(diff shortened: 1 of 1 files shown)


YRWoods:
Withdrawing.

Issue #81, opened 2026-07-06
[SECURITY][HIGH] Swagger UI / OpenAPI spec served without authentication — full API blueprint exposed


LucasMaupin:
## Summary
The Fastify API key authentication guard only fires when `req.url.startsWith('/api/v1')`. The Swagger UI (`/documentation`) and its machine-readable specs (`/documentation/json`, `/documentation/yaml`) fall outside this prefix and are served unauthenticated even when `API_KEY` is configured. The spec exposes full route signatures, request/response schemas, and bearer-auth configuration — a significant reconnaissance aid for attackers targeting the system.

## CVSS Score
7.5 (High) — CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:N/A:N

## Category
OWASP A01 — Broken Access Control

## Affected File(s)
- `src/server.ts:92-95` (auth guard URL prefix check)
- `src/server.ts:115` (Swagger registration)

## Steps to Reproduce
1. Configure and start the backend with `API_KEY=secret`
2. `curl http://localhost:3000/documentation/json` — returns the full OpenAPI spec without any authentication header

## Remediation
Extend the auth hook to also protect `/documentation`:
```typescript
fastify.addHook('onRequest', async (req, reply) => {
  const exempt = ['/api/v1/status', '/api/v1/reconnect']; // existing exemptions
  if (!req.url.startsWith('/api/v1') && !req.url.startsWith('/documentation')) return;
  // apply API key check
});
```
Alternatively, serve Swagger only in non-production environments by gating the registration on `NODE_ENV !== 'production'`.

## References
- [OWASP A01:2021 — Broken Access Control](https://owasp.org/Top10/A01_2021-Broken_Access_Control/)
- [Fastify docs — Route Authentication](https://fastify.dev/docs/latest/Reference/Hooks/)

LucasMaupin:
The Swagger UI and OpenAPI spec endpoints fall outside the API key auth guard, so they're accessible without any credentials. The scope is clear: extend the auth hook prefix in `server.ts` to cover `/documentation`. Labelled `bug` and `ready-for-dev`. The security-engineer will follow up on priority.

LucasMaupin:
PR opened: https://github.com/Eyevinn/open-live/pull/91

birme:
🤖 Picked up by daily-backlog-pr (backend-developer). Starting implementation.

birme:
[...]

PR #123, opened 2026-07-15
fix(security): gate /ws with API_KEY and accept ?token=

YRWoods:
## Summary
Companion to open-live-studio#38 / #81 family.

When `API_KEY` is set, require auth for `/ws/*` and `/documentation*`. Accept `Authorization: Bearer`, `?key=`, or `?token=`.

## Test plan
- [x] `npm run typecheck`
- [ ] WS upgrade without key when API_KEY set → 401
- [ ] `?token=` or `?key=` matching API_KEY → connects


YRWoods:
Withdrawing.

PR #140, opened 2026-07-28
fix: enable trustProxy and use req.ip for rate limiting and audit log (closes #139)

### LINKED ISSUE #139: [SECURITY][MEDIUM] X-Forwarded-For blindly trusted without trustProxy — rate limit bypass
## Summary

The rate-limit `keyGenerator` at `src/server.ts:72` reads the client IP from the `X-Forwarded-For` header without Fastify's `trustProxy` option being enabled. When `trustProxy` is not set, Fastify does not validate that `X-Forwarded-For` originates from a trusted proxy, so any client can spoof this header to bypass rate limiting by rotating fake IPs on every request. The same header is used in the audit log at `src/server.ts:133`, so spoofed IPs pollute the audit trail.

## CVSS Score
5.3 (Medium) — CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:L/A:L

## Category
OWASP A05 — Security Misconfiguration

## Affected File(s)
- `src/server.ts:72` (keyGenerator)
- `src/server.ts:133` (audit log IP extraction)

## Steps to Reproduce
1. Start the server without `trustProxy: true`.
2. Send requests with a rotating `X-Forwarded-For` header:
   ```bash
   for i in $(seq 1 500); do
     curl -H "X-Forwarded-For: $i.0.0.1" http://localhost:3000/api/v1/productions
   done
   ```
3. All 500 requests succeed; the rate limit counter never reaches the threshold for any single IP.

## Remediation
Add `trustProxy: true` (or the specific proxy CIDR) to the Fastify constructor, and only use the resolved `req.ip` (which Fastify sets correctly when `trustProxy` is configured) in the keyGenerator and audit log:

```typescript
const server = Fastify({
  trustProxy: true,   // <-- add this
  logger: false,
});

// keyGenerator — use req.ip directly; Fastify resolves it from the trusted X-Forwarded-For chain
keyGenerator: (req) => req.ip,
```

If the server is not behind a proxy, remove the `X-Forwarded-For` parsing entirely and use `req.ip` (TCP source IP) for both rate limiting and audit logging.

## References
- [Fastify trustProxy documentation](https://fastify.dev/docs/latest/Reference/Server/#trustproxy)
- [OWASP: Rate Limiting Best Practices](https://owasp.org/www-community/controls/Blocking_Brute_Force_Attacks)

LucasMaupin:
## Summary

- Add `trustProxy: true` to the Fastify constructor so the framework validates the `X-Forwarded-For` chain and resolves `req.ip` to the real client IP
- Replace the manual `x-forwarded-for` header extraction in the rate-limit `keyGenerator` with `req.ip` — Fastify now handles the chain correctly
- Replace the same pattern in the audit-log `onResponse` hook with `req.ip`

Without `trustProxy`, any client could send a spoofed `X-Forwarded-For` header to make the rate limiter count each request against a different fictitious IP, rendering the limit ineffective. Spoofed IPs also polluted the audit log.

Closes #139

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -37,6 +37,11 @@ export async function buildServer() {
     disableRequestLogging: true,
     // Prevent memory exhaustion via oversized request bodies (1 MB limit)
     bodyLimit: 1_048_576,
+    // Trust the X-Forwarded-For header from the ingress proxy so that req.ip
+    // resolves to the real client IP rather than the proxy's address.
+    // Without this, the header is treated as user-controlled input, allowing
+    // spoofed IPs to bypass rate limiting.
+    trustProxy: true,
   });
 
   // CORS must be registered before Helmet so its onRequest hook runs first
@@ -69,7 +74,7 @@ export async function buildServer() {
     // Skip health/ready probes — they are high-frequency and come from the cluster
     allowList: (req: { url: string }) => req.url === '/health' || req.url === '/ready',
     skipOnError: false,
-    keyGenerator: (req: { headers: Record<string, string | string[] | undefined>; ip: string }) => (req.headers['x-forwarded-for'] as string ?? req.ip).split(',')[0]!.trim(),
+    keyGenerator: (req: { ip: string }) => req.ip,
     errorResponseBuilder: (_req, context) => ({
       error: 'Too many requests',
       statusCode: 429,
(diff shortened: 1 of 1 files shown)


birme (commented):
APPROVE — trustProxy is a valid Fastify constructor option; switching keyGenerator and the audit hook from raw x-forwarded-for reads to req.ip is correct and closes the spoofing bypass. Both call sites in server.ts were updated and the commit has a Co-Authored-By trailer.

Issue #138, opened 2026-07-27
[SECURITY][MEDIUM] Timing-unsafe API key comparison — brute-force enumeration side-channel


LucasMaupin:
## Summary

The API key is compared using JavaScript's native `!==` operator at `src/server.ts:123`. V8 short-circuits string comparison on the first mismatched character, creating a measurable timing side-channel. In co-hosted or containerised environments (shared Kubernetes nodes), an attacker can statistically enumerate the key one character at a time by measuring response latencies over many requests.

## CVSS Score
5.3 (Medium) — CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:N/A:N

## Category
OWASP A07 — Identification and Authentication Failures

## Affected File(s)
- `src/server.ts:123`

## Steps to Reproduce
1. Set `API_KEY` on the backend.
2. From a co-located client, send requests varying the first character of the key.
3. Measure response latency over 10,000+ samples.
4. Statistical analysis reveals matching prefix length.

## Remediation
Replace the comparison with Node.js `crypto.timingSafeEqual`:

```typescript
import { timingSafeEqual } from 'crypto';

const a = Buffer.from(provided ?? '');
const b = Buffer.from(config.apiKey);
if (a.length !== b.length || !timingSafeEqual(a, b)) {
  reply.code(401).send({ error: 'Unauthorized' });
  return;
}
```

## References
- [Node.js crypto.timingSafeEqual](https://nodejs.org/api/crypto.html#cryptotimingsafeequala-b)
- [OWASP: Timing Attacks](https://owasp.org/www-community/attacks/Timing_attack)

LucasMaupin:
The API key check at `src/server.ts:123` uses `!==`, which short-circuits on the first mismatched byte — this leaks timing information that can be exploited statistically from a co-located process to enumerate the key one character at a time.

Labelled `bug` + `priority: P2` + `ready-for-dev`. The fix is a 3-line swap to `crypto.timingSafeEqual` — the issue includes a ready-to-use code snippet. Marked `good-first-issue`; a contributor should be able to pick this up cleanly.

LucasMaupin:
PR opened to fix this: https://github.com/Eyevinn/open-live/pull/141

The fix replaces the `!==` comparison with `crypto.timingSafeEqual` — both values are converted to `Buffer` first (with a length pre-check), then compared in constant time so the timing side-channel is eliminated.

PR #141, opened 2026-07-28
fix: use timingSafeEqual for API key comparison to close timing side-channel (closes #138)

LucasMaupin:
## Summary

- Import `timingSafeEqual` from Node.js built-in `crypto`
- Replace the `provided !== config.apiKey` comparison with a constant-time check: convert both strings to `Buffer`, verify equal length first, then call `timingSafeEqual`

The native `!==` operator allows V8 to short-circuit on the first mismatched character, creating a measurable latency difference that a co-located attacker can exploit to enumerate the API key one character at a time. `timingSafeEqual` compares all bytes in constant time regardless of where the first mismatch occurs.

Closes #138

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #142, opened 2026-07-28
fix(security): timing-safe API key comparison and trustProxy (#138, #139)

1devananddewangan234:
Fixes two security vulnerabilities:\n\n#138 - Timing-unsafe API key comparison: replaced !== with crypto.timingSafeEqual using 256-byte padded buffers\n\n#139 - X-Forwarded-For blindly trusted: added trustProxy: true to Fastify, changed keyGenerator and audit log to use req.ip\n\nFixes #138, #139

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -5,6 +5,7 @@ import rateLimit from '@fastify/rate-limit';
 import websocket from '@fastify/websocket';
 import swagger from '@fastify/swagger';
 import swaggerUi from '@fastify/swagger-ui';
+import { timingSafeEqual } from 'crypto';
 import { ZodError } from 'zod';
 import { config } from './config.js';
 import { isDbConnected } from './db/index.js';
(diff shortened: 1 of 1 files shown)


birme:
Closing — this branch was cut before #140 merged and, if merged now, would silently revert three pieces of already-shipped hardening it doesn't mention: the /ws/ prefix in the auth-exemption check (server.ts) — removing it reopens an unauthenticated WebSocket production-control hole, more severe than either #138 or #139 this PR claims to fix; crossOriginResourcePolicy same-origin to cross-origin; and strictPreflight true to false. #139 is already fixed and closed via #140. The timingSafeEqual technique here (fixed-size padded buffer) is a nice touch and #141 (open, addressing #138) is welcome to borrow it, but this PR as submitted must not be merged. Thanks for the report though — happy to see a fixed-up version scoped to just #138.

PR #155, opened 2026-08-15
fix: add silent audiotestsrc per test pattern source to unblock audio mixer (closes #154)

### LINKED ISSUE #154: Production with only test pattern sources fails to activate — audio mixer channel with no audio source in the Strom graph
## Summary

A production that contains **only test pattern sources** (`test1` / `test2`) does not work — activation produces a Strom flow that never comes up properly. A quick investigation points at the **missing audio** in the generated Strom graph as the root cause: test patterns are video-only, so the audio mixer ends up with an input channel that nothing is ever linked to.

## Steps to reproduce

1. Create a new production.
2. Add two sources, both test patterns (e.g. "Test - Pinwheel" and "Test - Colors"), and nothing else.
3. Activate the production.

**Expected:** the flow starts and both test patterns are previewable/mixable, with silence on the programme audio.

**Actual:** it does not work — the flow does not come up as expected.

## Analysis

In `src/lib/flow-generator.ts`:

* Test pattern sources are generated as a bare `videotestsrc` element and are linked to the **video** mixer only. They contribute no audio branch and no link into the audio mixer — the code explicitly notes that "test1/test2 sources don't carry audio".
* The audio mixer channel count is computed from the number of audio-bearing sources:

  ```ts
  const srtEfpCount = sortedAssignments.filter(/* streamType !== test1 && !== test2 */).length;
  const numChannels = Math.max(1, srtEfpCount);
  props['num_channels'] = numChannels;
  ```

With a test-pattern-only production `srtEfpCount === 0`, so the `Math.max(1, …)` floor forces the audio mixer to be built with **one channel that has no producer linked to it**. An audio mixer/aggregator will wait for buffers on every configured sink pad, so an unconnected channel stalls the audio branch — and with it the WHEP output, which is wired to `audioMixer:main_out` (and `monitor_out`).

So the graph is structurally invalid for this case: audio mixer channels > 0, audio sources = 0.

## Suggested fix (pick one)

1. **Give test patterns silent audio.** Add an `audiotestsrc` (`wave=silence`) branch per test pattern source, so every source is audio-bearing and the existing channel accounting just works. This keeps the graph uniform and is probably the smallest conceptual change.
2. **Handle the zero-audio-source case explicitly.** If there are no audio-bearing sources, either omit the audio mixer and its links entirely (and drop `num_audio_tracks` on the WHEP output accordingly), or feed the one forced channel with a silence generator.

Option 1 also fixes the mixed case where a test pattern sits alongside real sources but occupies no audio channel, which makes channel indices and mixer strips inconsistent with the source list.

## Notes

* Repro is trivial and does not need any external SRT/WHIP feed, so this is a good smoke-test candidate: "activate a production with only test patterns" should be covered by a test.

LucasMaupin:
## Summary

- Adds an `audiotestsrc` (`wave=silence`, `is-live=true`) branch per test pattern source (`test1`/`test2`), wired through a `time_offset` block into both the vision mixer's `audio_in_N` pad and the audio mixer's `input_N` channel.
- Updates the `num_channels` computation on the audio mixer to count test patterns as audio-bearing sources, keeping channel indices consistent with the source list.
- Adds a dedicated `flow-generator.test.ts` covering: test-pattern-only production (2× audiotestsrc, correct `num_channels`, links to mixer), mixed virtual-source production, and single test pattern resolving without throwing.

## Root cause

Test pattern sources produced no audio branch. `num_channels` on the audio mixer was `Math.max(1, 0) = 1`, but nothing was linked to that channel — the aggregator stalled waiting for buffers, preventing the flow from coming up.

Closes #154

### DIFF
--- a/src/__tests__/flow-generator.test.ts
+++ b/src/__tests__/flow-generator.test.ts
@@ -1,0 +1,141 @@
+/**
+ * Unit tests for activateStromFlow in flow-generator.ts.
+ *
+ * Strom client and CouchDB are mocked — no real services required.
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// ---------------------------------------------------------------------------
+// Mock CouchDB
+// ---------------------------------------------------------------------------
+
+vi.mock('../db/index.js', () => ({
+  getSourcesDb: () => ({ get: vi.fn().mockRejectedValue(new Error('not found')) }),
+  getGraphicsDb: () => ({ get: vi.fn().mockRejectedValue(new Error('not found')) }),
+}));
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function makeStromClient() {
+  const capturedFlows: Record<string, unknown>[] = [];
+  const client = {
+    flows: {
+      create: vi.fn().mockImplementation((flow: Record<string, unknown>) => {
+        capturedFlows.push(flow);
+        return Promise.resolve({ flow: { id: 'flow-test-123' } });
+      }),
+      start: vi.fn().mockResolvedValue({}),
+      delete: vi.fn().mockResolvedValue({}),
+    },
+    capturedFlows,
+  };
+  return client;
+}
+
+function makeProduction(sources: Array<{ sourceId: string; mixerInput: string }>, values?: Record<string, unknown>) {
+  return {
+    _id: 'prod-test-only',
+    _rev: '1-abc',
+    type: 'production',
+    name: 'Test Production',
+    status: 'inactive',
+    sources,
+    graphicAssignments: [],
+    values: values ?? {},
+    pipeline: { stromConfig: null, status: 'stopped' },
+    graphics: [],
+    macros: [],
+    tally: { pgm: null, pvw: null },
+    createdAt: '2026-01-01T00:00:00.000Z',
+    updatedAt: '2026-01-01T00:00:00.000Z',
+  };
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe('activateStromFlow — test pattern sources', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('adds one audiotestsrc per test pattern and sets num_channels correctly (test-only production)', async () => {
+    const { activateStromFlow } = await import('../lib/flow-generator.js');
+    const strom = makeStromClient();
+
+    const production = makeProduction([
+      { sourceId: '__test1__', mixerInput: 'video_in_1' },
+      { sourceId: '__test2__', mixerInput: 'video_in_2' },
+    ]);
+
+    await activateStromFlow(production as never, strom as never);
+
+    expect(strom.flows.create).toHaveBeenCalledOnce();
+    const flow = strom.capturedFlows[0]!;
+    const elements = flow['elements'] as Array<Record<string, unknown>>;
+    const blocks = flow['blocks'] as Array<Record<string, unknown>>;
+    const links = flow['links'] as Array<Record<string, unknown>>;
+
+    // Each test pattern should have one audiotestsrc element
+    const audioTestSrcs = elements.filter((e) => e['element_type'] === 'audiotestsrc');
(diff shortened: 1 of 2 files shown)


birme (commented):
**pr-reviewer verdict: APPROVE**

Correctness: The fix correctly addresses the stalled-aggregator root cause. Each test1/test2 source now gets a silent `audiotestsrc` (wave=silence, is-live=true) wired through a `time_offset` block into both the vision mixer `audio_in` pad and the audio mixer `input_` channel, so no audio mixer channel is left without a linked producer. The `num_channels` computation is updated to count test patterns as audio-bearing (`src != null`), keeping channel indices consistent with the newly-added audio branches. `audioChannelIndex` is incremented in the same pattern as the existing `html` branch, so indices stay in sync.

Tests: Three meaningful vitest cases with Strom/CouchDB mocked — assert one audiotestsrc per test pattern, `wave === 'silence'`, correct `num_channels` (test-only and mixed-source), links into the audio mixer, and successful resolution for a test-only production. Good coverage of the regression.

Security: No secrets, no logging concerns; behavioral topology change only.
Conventions: Branch from main, imperative commit, `Co-Authored-By` trailer present.

No blocking issues. Approving as a comment (caller controls merge).

birme (commented):
## pr-reviewer review — PR #155 (closes #154)

Reviewed as the **pr-reviewer** agent (read-only; comment only, no approve/merge).

### 1. Guesswork / fabrication check
Verified every field, block id, and API call in the diff against the actual repo at both `main` and the PR head (`fix/issue-154-test-pattern-silent-audio`):
- `audiotestsrc` element with array `position` and a `properties` object — matches the existing `videotestsrc`/`cefsrc`/`cefdemux` element pattern already used in `flow-generator.ts`. `wave`/`is-live` are standard GStreamer `audiotestsrc` props; the codebase already emits raw GStreamer `element_type`s.
- `builtin.time_offset` block, `${block}:audio_in_${padIndex}` vision-mixer pad, and `${audioMixerBlockId}:input_${audioChannel + 1}` channel — identical to the existing html/whip/srt branches (lines ~522/554/589 on `main`). No new API surface invented.
- `ch{N}_label` label prop, `builtin.mixer` block id, and the `{ id, block_definition_id, name, properties, position }` block shape all pre-exist.
- Test file: `activateStromFlow(production, strom)` signature, `created.flow.id` return (`{ flowId }`), `../db/index.js` `getSourcesDb`/`getGraphicsDb`, and `strom.flows.create/start/delete` all match real code. Mock shapes are accurate.

No `FABRICATED` findings.

### 2. Correctness — does it fix the bug?
Yes. Root cause (audio mixer channel with no linked producer) is addressed: each `test1`/`test2` source now emits a silent `audiotestsrc` → `time_offset` → `audioMixer:input_N`, so every counted channel has a real producer feeding it. The `num_channels` change (`src != null` instead of excluding test1/test2) now counts test patterns, and this stays consistent with the per-source `audioChannelIndex++` in the loop — every non-null source hits exactly one branch (test/html/whip/else), so channel count == linked count. Deleted real sources resolve to `null` and are correctly excluded from both the count and the loop (`if (!source) continue`). Invariant holds.

### 3. Mandatory rules (CLAUDE.md)
- Branch off `main` (`fix/issue-154-...`), base `main` — no direct push to main. PASS.
- Co-author trailer present: `Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>`. PASS.
- No hardcoded SRT ports introduced — the only SRT URI is the pre-existing untouched fallback. PASS.
- Watchdogs: not applicable to a local silent live source; network SRT inputs are unchanged. PASS.

### 4. Security & tests
- No security-sensitive surface (no secrets, no external I/O added).
- New `flow-generator.test.ts` covers test-only production (2× audiotestsrc, num_channels=2, links into mixer), mixed virtual sources (num_channels=3), and single-pattern resolution without throwing. Good coverage of the new behavior.

### Verdict: APPROVE
Fix is correct, consistent with existing patterns, no fabricated APIs, mandatory rules satisfied, and well-tested. (Posting as a comment only per pr-reviewer role — orchestrator handles the actual approve/merge.)

birme (approved):
Approved by daily-backlog-pr pr-reviewer gate (verdict APPROVE, verified against source). Merging.

PR #156, opened 2026-08-20
feat(ws): broadcast the PGM background input behind a PiP

wagenet:
_Generated by Claude Code with human review._

While a PiP occupies PGM, `tally.pgm` is null and the real background input lives only in the module-level `pgmBgByProduction` map, which no `broadcast(...)` ever reads. Any client recording tally therefore stores "no source on program" for the whole duration of every PiP — wrong in the worst direction, because it looks like data rather than absence.

`PIP_STATE` does not close the gap: `pgmPip` is a slot index and `pips[i].bg` is the PiP configuration's own background, neither of which is the mixer input the PiP is composited over. Connect is the case no replay reconstructs — a client attaching mid-show gets `TALLY {pgm: null, …}` and `PIP_STATE {pgmPip: N}`, with nothing carrying the background.

## What changed

`feat(ws): broadcast the PGM background input behind a PiP` adds `pgmBg` to all nine places the server emits `TALLY` — eight `broadcast()` calls plus the connect-time `socket.send` — so a consumer can distinguish three states rather than two:

| `pgmPip` | `pgmBg` | Meaning |
| --- | --- | --- |
| null | null | No PiP on program |
| set | set | PiP over a known input |
| set | null | PiP over nothing (empty PGM and empty PVW at take time) |

It rides on `TALLY` rather than `PIP_STATE` because `pgm: null` is the ambiguous field and it lives on `TALLY`, while `PIP_STATE` is emitted conditionally — TRANSITION only when a PiP was on program, CUT only when a PiP was on program *or* in preview, the three macro paths never — so a field carried there would go stale across exactly those transitions. That commit is additive: each payload gains one field, nothing else changes value, and no mixer state is touched.

On the take that moves a PiP to PGM the value falls back to the outgoing PGM input when no PVW source was displaced:

```js
const newPgmBg = newPgmPip !== null ? (pvwBeforePip ?? tally.pgm) : null;
```

matching the `to_input` the Strom transition computes a few lines below. `pvwBeforePip` is null whenever a PiP is selected into an empty preview, ordinary at the top of a show; without the fallback the broadcast reports no background while the mixer composites over the previous program input.

`fix(ws): record the PGM background before the Strom calls` moves the `pgmBgByProduction` write above the broadcast. The TAKE handler derives the broadcast value from `pvwBeforePipByProduction` but wrote the map inside `if (doc.stromFlowId && doc.mixerBlockId)`, after the Strom calls succeed, and the connect handler reads that map. With Strom unconfigured, or if `strom.mixer.transition` throws into its `catch`, a client attaching during a PiP is sent `pgmBg: null` — precisely the case the field exists for.

`pvwBeforePipByProduction` is now read once and reused for both the broadcast value and the transition's `to_input`. The two reads sat either side of `await makeStromClient()`, so a CUT or TRANSITION interleaving at that await could change the map between them. The `to_input` expression is unchanged; only its operand is now the captured local.

The remaining two commits document `pgmBg` in `docs/controller-websocket.md` and tighten the comments added above.

## Decision needed

When the take displaced no PVW source, the map now holds the outgoing PGM input rather than `null`, so a later CUT while the PiP is still on PGM passes that input as `from_input` instead of letting it collapse to `toIndex` at `controller.ts:254`. That is a real change to a live mixer parameter, not just a reporting field.

I read it as a fix: `pgmPipByProduction` is already set unconditionally above, so the handler is committed to the take regardless of what Strom did, and `null` was never more accurate than the real input. @birme read it the same way on review. Happy to drop that commit and accept the connect-time divergence instead if you would rather the map keep its current value.

## A gap this does not close

The three macro paths (`controller.ts:991`, `:1000`, `:1007`) update `tally` but never touch `pgmPipByProduction` or `pgmBgByProduction` and never emit `PIP_STATE`, so after a macro CUT over a PiP the new field reports the old background alongside a real `pgm` — a fourth combination the table does not cover.

That is pre-existing rather than introduced here, and larger than the reporting symptom: those paths also never restore the PiP into preview the way the interactive CUT does, which is what actually reaches the mixer. #179 and #180 fix it. The table should not be read as exhaustive until they land.

## Testing

- `npx vitest run` — 120/120 passing, including the "pgmBg with no Strom flow configured" case that ships with commit 2. Reverting commit 2 makes it fail, reporting `pgmBg: null` where the take broadcast `video_in_1`.
- `npx tsc --noEmit` clean.
- Verified against a live production driven through ordinary cuts and a PiP take: `pgmBg` appears in the `TALLY` broadcast and matches what was on air.

Not exercised at runtime: the `pvwBeforePip ?? tally.pgm` fallback. It is derived from the adjacent `to_input` computation rather than observed, so the reasoning is worth a look.

### DIFF
--- a/docs/controller-websocket.md
+++ b/docs/controller-websocket.md
@@ -108,7 +108,7 @@ are emitted from `src/ws/controller.ts` and `src/services/meter-relay.ts`:
 
 | `type` | Fields | Emitted when |
 |---|---|---|
-| `TALLY` | `pgm: string \| null`, `pvw: string \| null`, `transitionType?: string`, `durationMs?: number` | Tally (PGM/PVW) changes; also sent on connect |
+| `TALLY` | `pgm: string \| null`, `pvw: string \| null`, `pgmBg: string \| null`, `transitionType?: string`, `durationMs?: number` | Tally (PGM/PVW) changes; also sent on connect |
 | `PIP_STATE` | `pgmPip: number \| null`, `pvwPip: number \| null`, `pips: PipConfig[]` | PiP program/preview/config changes; also sent on connect |
 | `FTB_STATE` | `active: boolean` | Fade-to-black state changes |
 | `ON_AIR` | `value: boolean` | Production goes on/off air (`GO_LIVE` / `CUT_STREAM`) |
@@ -134,6 +134,12 @@ are emitted from `src/ws/controller.ts` and `src/services/meter-relay.ts`:
 | `METER_DATA` | `elementId: string`, `peak`, `rms` | Audio meter tick (relayed from Strom); `elementId` is `main`, `monitor`, `ch{N}`, `aux{N}`, or `grp{N}` |
 | `LOUDNESS_DATA` | `elementId: 'main'`, `momentary`, `shortterm`, `integrated`, `loudness_range`, `true_peak` | EBU R128 loudness tick (relayed from Strom) |
 | `ERROR` | `error: string` | An inbound frame was invalid or an operation failed (sent to originating socket) |
+
+`pgmBg` is the mixer input a PiP on program is composited over. It is `null` unless
+`PIP_STATE.pgmPip` is set, so the two fields together distinguish an empty program
+(`pgmPip` null) from a PiP over a known input (both set) from a PiP over nothing
+(`pgmPip` set, `pgmBg` null). It is not tracked across the `MACRO_EXEC` cut,
+transition, and take paths, which leave it holding the value from before the macro.
 
 ### Connect-time snapshot
 
--- a/src/__tests__/ws-macro-pip.test.ts
+++ b/src/__tests__/ws-macro-pip.test.ts
@@ -164,6 +164,31 @@ beforeEach(() => {
 });
 
 // ---------------------------------------------------------------------------
+// The PGM background must be recorded even when Strom is unconfigured
+// ---------------------------------------------------------------------------
+
+describe('pgmBg with no Strom flow configured', () => {
+  it('records the background behind the PiP so later tallies still carry it', async () => {
+    mockGet.mockResolvedValue({
+      ...makeProductionDoc([{ type: 'CUT', sourceId: 'cam3' }]),
+      stromFlowId: undefined,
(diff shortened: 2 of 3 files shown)


birme:
Reviewed by actually running it: merged this branch onto current main (post-#182), resolved the 8 conflicting hunks by hand — all against #175's `persistMixerMutation()` refactor, which landed after this branch was cut; kept the durable-write helper and layered pgmBg onto its broadcast in each case. `npx tsc --noEmit` clean, `npx vitest run` — 107/107 passing (up from 106 after #182), including commit 3's new 'pgmBg with no Strom flow configured' case.

The core addition (adding pgmBg to every TALLY broadcast, including connect-time sync) is correct and exactly as described — verified pgmBgOf(), the nine broadcast sites, and that pgmBgByProduction/pvwBeforePipByProduction are read/written the way the commit messages claim. Commit 3's reordering (recording pgmBgByProduction before the Strom calls rather than inside the success branch) is also correct and closes a real bug: previously an unconfigured-Strom or thrown-transition build would broadcast a live TALLY with the right pgmBg but serve a *different* (stale/null) value to a client connecting afterward, since connect reads the map directly.

On the one open question in the PR description — whether the `pvwBeforePip ?? tally.pgm` fallback should also change what a *later* CUT's `from_input` resolves to when a PiP took an empty preview: I read this as correct, not just consistent. Before this PR, a CUT after such a take would let `from_input` collapse to `to_index` (i.e., cut into itself) purely because the map happened to hold `null" for reasons unrelated to Strom state — that's a coincidental collapse, not a deliberate invariant. Making the map hold the real input the mixer is actually compositing over is the more correct behavior for both the broadcast and the from_input, and it isn't exercised by any existing test either way, so nothing regresses.

This is still marked draft, so I'm not merging it — flagging back to the team since it changes real Strom transition parameters (from_input on a live mixer), not just a reporting field, and the author explicitly asked for a maintainer call on it. Recommend marking ready and merging once someone signs off on that read.

wagenet (commented):
Looks good to me!

wagenet:
@birme I've rebased it at taken it out of draft. Thanks for taking a look!

PR #157, opened 2026-08-20
fix(ws): record the PGM background before the Strom calls, not inside them

wagenet:
**Stacked on #156.** That PR adds the `pgmBg` field; this one makes it correct on connect. The diff below contains both commits until #156 merges — only the second, `fix(ws): record the PGM background before the Strom calls`, belongs to this PR.

## Problem

#156 adds `pgmBg` to every `TALLY` broadcast. Every live broadcast is correct. The connect-time sync — the case the field exists for — is not:

```
TAKE broadcast:  {"type":"TALLY","pgm":null,...,"pgmBg":"video_in_1"}
on connect:      {"type":"TALLY","pgm":null,...,"pgmBg":null}
                 {"type":"PIP_STATE","pgmPip":0,...}
```

`pgmPip` says a PiP is on program while `pgmBg` says there is no background. The server contradicts the tally it broadcast moments earlier.

The TAKE handler derives the broadcast value from `pvwBeforePipByProduction`, but only writes `pgmBgByProduction` inside `if (doc.stromFlowId && doc.mixerBlockId)`, after those calls succeed. The connect handler reads that map. So with Strom unconfigured — how most people will run a dev instance — or if `strom.mixer.transition` throws into its `catch`, a client attaching during a PiP is sent `pgmBg: null`.

That is precisely the case a subscriber cannot reconstruct, because the value is never re-broadcast, which was the entire justification for the field.

## Why the old placement was not a bug

It was correct for the map's original job, and its comment said so — track the background *"so CUT/TRANSITION can pass the correct `from_input`"*. If Strom never took the transition, there is no Strom-side background worth remembering for the next `from_input`.

#156 gives the map a second job: it is now also what connecting clients are told. The old placement does not satisfy the new one. This PR moves the assignment above the broadcast, derived from the same in-scope value, so the map and the wire cannot disagree.

## Risk

**`pvwBeforePipByProduction` is now read once and reused** for both the broadcast value and the transition's `to_input`, rather than read a second time inside the Strom block. The two reads were separated by `await makeStromClient()`, so a CUT or TRANSITION interleaving at that await could change the map between them and let the reported background disagree with the input the mixer actually cut to — the same class of divergence this PR closes.

The `to_input` expression itself is unchanged; only its operand is now the captured local, which holds the same value in the single-client case.

**One behavioural change, disclosed deliberately.** The value the map ends up holding differs in a single case: when the take displaced no PVW source, it now holds the outgoing PGM input rather than `null` — matching what the mixer actually composites over, and what the broadcast reports. A later CUT while the PiP is still on PGM therefore passes that input as `from_input` instead of letting it collapse to `to_index` at `controller.ts:162`.

I believe that is a fix rather than a regression: `pgmPipByProduction` is already set unconditionally above, so the handler is committed to the take having happened regardless of what Strom did, and `null` there was never more accurate than the real input. But it is a change on a live transition path, and it is the reason this is a separate PR rather than folded into #156 — happy to drop it if you would rather the map keep its current value and accept the divergence.

## Testing

- Verified against a live production: before this change the connect-time tally reported `pgmBg: null` during a PiP; after it reports the background input.
- `npx tsc --noEmit` clean.

### DIFF
--- a/src/ws/controller.ts
+++ b/src/ws/controller.ts
@@ -312,6 +312,20 @@ const pvwBeforePipByProduction = new Map<string, string | null>()
  */
 const pgmBgByProduction = new Map<string, string | null>()
 
+/**
+ * The real input sitting behind a PiP that is on program, or null.
+ *
+ * While a PiP occupies PGM, `tally.pgm` is null and the background input is
+ * tracked only in `pgmBgByProduction`, which is never otherwise sent to a
+ * subscriber. Without it a client cannot tell "nothing on program" from "a PiP
+ * over input 3" — including on connect, where the distinction is otherwise
+ * unrecoverable, because the value is never re-broadcast.
+ *
+ * Read-only: nothing here changes mixer state or what airs.
+ */
+const pgmBgOf = (productionId: string): string | null =>
+  pgmBgByProduction.get(productionId) ?? null
+
 
 /** Wipe all per-production audio state. Called when the pipeline changes or production deactivates. */
 export function clearAudioState(productionId: string): void {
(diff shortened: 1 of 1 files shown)


wagenet:
Folded into #156 as the second of two commits, since the field this adds is not fully correct without it — #156 alone reports `pgmBg: null` on connect whenever Strom is unconfigured or the transition throws. Keeping them in one PR avoids landing the feature with its motivating case still broken. Same two commits, unchanged; the reasoning that was here is now in #156's description under commit 2.

Issue #149, opened 2026-08-10
[SECURITY][LOW] CouchDB all document types share one database and one nano handle — no collection-level isolation


LucasMaupin:
## Summary

All document types (productions, sources, graphics, outputs, configs) are stored in a single CouchDB database \`open-live\` and accessed through a single nano \`DocumentScope\` that is cast to different TypeScript types via \`as unknown as Nano.DocumentScope<T>\`. The \`getSourcesDb()\`, \`getGraphicsDb()\`, etc. functions simply return the same handle with a different type cast. CouchDB's \`db.find()\` with \`selector: { type: '...' }\` queries work correctly at runtime, but there is no database-level isolation between document types. A bug in any one route could accidentally read, overwrite, or delete documents of a different type. The document \`_id\` prefix conventions (\`prod-\`, \`src-\`, \`gfx-\`, \`out-\`, etc.) are the only guard.

## CVSS Score
3.1 (Low) — CVSS:3.1/AV:N/AC:H/PR:H/UI:N/S:U/C:L/I:L/A:N

## Category
OWASP A04 — Insecure Design

## Affected File(s)
- \`src/db/index.ts:1-41\`

## Steps to Reproduce
This is a design issue rather than a directly exploitable attack. A code path calling \`getProductionsDb().get('src-<uuid>')\` would silently return a \`SourceDoc\` through the \`ProductionDoc\` type handle without TypeScript or CouchDB catching it at runtime.

## Remediation
Consider separating document types into distinct CouchDB databases (e.g., \`open-live-productions\`, \`open-live-sources\`, \`open-live-graphics\`). This adds database-level access isolation, simplifies per-collection backup/restore, and prevents cross-type document operations. As a short-term mitigation, add runtime \`doc.type\` assertions in \`getDb().get()\` paths before using the result.

## References
- [CouchDB Security Documentation](https://docs.couchdb.org/en/stable/intro/security.html)
- OWASP A04:2021 — Insecure Design

birme:
🤖 **Triaged by daily-backlog-pr (architect)** → **Ready**. Classification: **single-service** (bug). Implement the short-term mitigation (runtime `doc.type` assertions in `getDb().get()` paths); the long-term DB-split is a separate future item. Phase 2 plain single-service path (backend-developer).

PR #165, opened 2026-08-22
fix(db): assert doc type on cross-collection reads (#149)

birme:
## Summary
- Fixes **#149** ([SECURITY][LOW] — A04 Insecure Design): all document types share one CouchDB database and one nano handle re-typed per collection via `as unknown as Nano.DocumentScope<T>`. A code path reading the wrong collection (e.g. `getDb().get('src-<uuid>')`) silently returned a mismatched document — the `_id` prefix conventions were the only guard.

## Changes
- `src/db/index.ts`: add `withTypeGuard()` wrapping each collection handle's `.get()`. After a fetch it asserts the returned document's `type` discriminator matches the collection it was requested from, throwing on a genuine cross-type mismatch. This is the short-term mitigation the issue proposes; the long-term DB-per-collection split remains a separate future item.

## Design notes
- **Non-breaking:** documents with no `type` field (legacy) are tolerated — only a present-and-mismatched `type` throws.
- **No type regression:** the wrapper is a `Proxy` returning the same `DocumentScope<T>` type, so call-site types are unchanged; only runtime behaviour is added.
- Centralized in `db/index.ts` rather than touching all 42 `.get()` call sites.

## Test Plan
- [x] `pnpm typecheck` passes
- [x] `vitest run`: no new failures vs baseline `main` (both are 27 failed/1 passed — the suite requires a live CouchDB/Strom not present in CI; this change introduces none).
- [ ] Manual: reading a `src-` id through `getDb()` (productions handle) now throws a clear type-mismatch error instead of returning a `SourceDoc`.

## Checklist
- [x] No hardcoded SRT ports
- [x] docs updated if needed (n/a)

🤖 Generated with Claude Code

### DIFF
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -5,8 +5,40 @@ import type { ProductionDoc, SourceDoc, ProductionConfigDoc, GraphicDoc, OutputD
 
 let db: Nano.DocumentScope<ProductionDoc>;
 
+// All document types share one physical CouchDB database and one nano handle,
+// which is re-typed per collection via `as unknown as ...`. There is no
+// database-level isolation, so a wrong-collection read (e.g. fetching a
+// `src-` doc through the productions handle) would otherwise return a
+// mismatched document silently. `withTypeGuard` wraps `.get()` to assert the
+// returned document's discriminator matches the collection it was fetched
+// from. It only throws on a genuine cross-type mismatch — documents with no
+// `type` field (legacy) are tolerated so existing data keeps working.
+function withTypeGuard<T extends { type?: string }>(
+  scope: Nano.DocumentScope<T>,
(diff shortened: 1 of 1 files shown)


birme (commented):
**pr-reviewer verdict: APPROVE** (independent agent review; implementer did not self-review).

Highest-risk item, verified carefully: all five `expectedType` strings match the `type` literals in `src/db/types.ts` (production/source/production-config/graphic/output). The Proxy intercepts only `.get()`; insert/find/destroy/list pass through via `Reflect.get`. Non-breaking (undefined `type` tolerated); no type regression (returns `DocumentScope<T>`); `pnpm typecheck` passes; no new vitest failures vs `main` baseline.

PR #172, opened 2026-08-24
fix(sources): encrypt SRT passphrases at rest with AES-256-GCM

### LINKED ISSUE #160: [SECURITY] Encrypt SRT passphrases at rest before storing in CouchDB
Split from #80 (breakdown by daily-backlog-pr / architect triage).

## Scope
SRT source addresses containing passphrases (`srt://host:port?passphrase=secret`) are stored verbatim as plaintext in CouchDB source documents (`src/routes/sources.ts`). Encrypt the passphrase portion before persisting.

## Remediation
- Encrypt the passphrase using AES-256-GCM with a key sourced from an environment variable (e.g. `SRT_PASSPHRASE_KEY`).
- Encrypt on write; decrypt on read before masking for API responses (`maskSrtPassphrase()`) and before passing to Strom.
- Alternatively, store only a reference ID and keep the passphrase in a dedicated secrets store (parameter store / Vault).

## Out of scope (tracked separately)
- CouchDB admin port exposure → #62
- Passphrase rotation procedure → companion sub-issue

Parent: #80

birme:
Closes #160

## Summary

SRT source addresses can embed a passphrase (`srt://host:port?passphrase=secret`). These were stored verbatim as plaintext in CouchDB source documents, exposing live-stream credentials to anyone with DB read access (backups, replication, operators). This PR encrypts the passphrase portion at rest using AES-256-GCM and decrypts it on read, so API masking and Strom activation are unchanged.

## Changes

- **New module `src/lib/srt-passphrase-crypto.ts`**
  - AES-256-GCM encrypt/decrypt of a single passphrase value.
  - Self-describing, url-safe, versioned wire format: `encv1:<base64url(iv | tag | ciphertext)>` (12-byte IV, 16-byte GCM auth tag, then ciphertext) — the IV and tag are stored so decrypt can parse them back.
  - Key loaded from `SRT_PASSPHRASE_KEY`, accepted as **base64 or hex**, validated to be exactly **32 bytes** (clear error otherwise).
  - **Fail closed in production**: when `NODE_ENV=production` and the key is unset, `loadKey()` throws. Outside production it degrades to a loud no-op (plaintext + `console.warn`) so local dev without a key still works.
  - Address-level helpers encrypt/decrypt **only** the `passphrase=` query param, leaving host/port/other params in cleartext so `srtUrl()` validation and operator masking keep working.
  - Never logs the plaintext passphrase or the raw key.
- **`src/routes/sources.ts`**
  - Encrypt on write in `POST` and `PATCH` (`encryptAddressPassphrase`).
  - Decrypt on read in `toApi()` before `maskSrtPassphrase()` so responses still return `passphrase=***`.
  - `PATCH` validation now runs against the decrypted (plaintext) address; stored value is re-encrypted only when the address actually changes.
- **`src/lib/flow-generator.ts`**: decrypt each source address right after loading from CouchDB so Strom receives a usable `srt_uri`. Only decrypt-on-read path that hands the URL to Strom.
- **Backward compatible on read**: un-prefixed (legacy plaintext) passphrases pass through decryption unchanged, so existing source docs keep working. All new writes/updates are encrypted.
- **`.env.example`**: documented the new `SRT_PASSPHRASE_KEY` var.

## New required-in-production env var

`SRT_PASSPHRASE_KEY` is a **new required-in-production** env var (AES-256, 32 bytes, base64 or hex; generate with `openssl rand -base64 32`). The server fails to start in production without it. **Ops/deploy follow-up**: provision the key in the parameter store / K8s secret before the next production deploy.

## Test Plan

New unit tests in `src/__tests__/srt-passphrase-crypto.test.ts` (vitest, matching the existing runner) cover:

- [x] round-trip encrypt → decrypt
- [x] fresh IV per encryption (ciphertexts differ) and no double-wrapping
- [x] tamper detection — flipped ciphertext byte and truncated bundle both throw (GCM auth)
- [x] legacy plaintext pass-through on read (value + full address)
- [x] key handling — base64/hex decode to the same 32-byte key, short/empty key throws, missing key fails closed in production and falls back in dev
- [x] address helper encrypts only the `passphrase=` param, leaves other params/host intact
- [x] API response still masks the decrypted passphrase as `passphrase=***`

Verification run locally:

- `pnpm install --frozen-lockfile` — OK
- `pnpm typecheck` (`tsc --noEmit`) — passes clean
- `npx vitest run src/__tests__/srt-passphrase-crypto.test.ts` — 21/21 pass

Note: the pre-existing `macros.test.ts` and `activation.test.ts` suites fail identically on a clean `main` (unrelated DB-mock/`isDbConnected` mismatch, no CouchDB/Strom in CI) — this PR adds no new failures.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -34,3 +34,9 @@ LOG_LEVEL=info
 # the correct host. Prevents X-Forwarded-Host injection when behind a reverse proxy.
 # If not set, the URL is derived from the incoming request (safe only with trustProxy configured).
 # PUBLIC_BASE_URL=https://your-open-live-instance.example.com
+
+# AES-256 key (32 bytes, base64 or hex) used to encrypt SRT source passphrases at
+# rest in CouchDB. REQUIRED in production (NODE_ENV=production) — the server fails
+# to start without it. Left unset in local dev, passphrases are stored in plaintext
+# with a startup warning. Generate one with: openssl rand -base64 32
+# SRT_PASSPHRASE_KEY=change-me-32-byte-base64-or-hex-key
--- a/src/__tests__/srt-passphrase-crypto.test.ts
+++ b/src/__tests__/srt-passphrase-crypto.test.ts
@@ -1,0 +1,174 @@
+/**
+ * Tests for SRT passphrase encryption-at-rest (issue #160).
+ *
+ * No CouchDB or Strom needed — this exercises the pure crypto module and the
+ * route-facing masking behaviour. A deterministic 32-byte test key is injected
+ * via process.env.SRT_PASSPHRASE_KEY; resetKeyCache() clears the module cache
+ * between env changes.
+ */
+
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import {
+  encryptPassphrase,
+  decryptPassphrase,
+  isEncrypted,
+  encryptAddressPassphrase,
+  decryptAddressPassphrase,
+  decodeKey,
+  loadKey,
+  resetKeyCache,
+} from '../lib/srt-passphrase-crypto.js';
+
+// 32 bytes of 0x01..0x20 — base64 form used as the default test key.
+const KEY_BYTES = Buffer.from(Array.from({ length: 32 }, (_, i) => i + 1));
+const KEY_B64 = KEY_BYTES.toString('base64');
+const KEY_HEX = KEY_BYTES.toString('hex');
+
+const SECRET = 'topsecretvalue';
+const SRT_ADDR = `srt://cam.example.com:9000?passphrase=${SECRET}&latency=200`;
+
+const originalEnv = { ...process.env };
+
+beforeEach(() => {
+  resetKeyCache();
+  process.env['SRT_PASSPHRASE_KEY'] = KEY_B64;
+  delete process.env['NODE_ENV'];
+});
+
+afterEach(() => {
+  process.env = { ...originalEnv };
+  resetKeyCache();
+  vi.restoreAllMocks();
+});
+
+describe('key loading', () => {
+  it('decodes a base64 key to 32 bytes', () => {
+    expect(decodeKey(KEY_B64)).toHaveLength(32);
+  });
+
+  it('decodes a hex key to 32 bytes', () => {
+    expect(decodeKey(KEY_HEX)).toHaveLength(32);
+  });
+
+  it('base64 and hex forms decode to the same key', () => {
+    expect(decodeKey(KEY_B64).equals(decodeKey(KEY_HEX))).toBe(true);
+  });
+
+  it('throws on a short key', () => {
+    expect(() => decodeKey(Buffer.from('too-short').toString('base64'))).toThrow(/32 bytes/);
+  });
+
+  it('throws on an empty key', () => {
+    expect(() => decodeKey('   ')).toThrow(/empty/);
+  });
+
+  it('fails closed in production when the key is unset', () => {
+    resetKeyCache();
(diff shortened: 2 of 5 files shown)


birme (commented):
**pr-reviewer verdict: APPROVE** (code) — but see MERGE-NOW operational gate below.

Crypto scheme review (AES-256-GCM):
- Fresh random 12-byte IV per encryption (`randomBytes(IV_BYTES)`) — no IV reuse. ✔
- 16-byte GCM auth tag captured (`getAuthTag`) and verified on decrypt (`setAuthTag`); tamper and truncation tests confirm decrypt throws. ✔
- Key length validated to exactly 32 bytes; hex (64-char) checked before base64 to avoid mis-decoding; empty/short keys throw with secret-free messages. ✔
- Plaintext passphrase and raw key are never logged; error/warn messages are value-free. ✔
- Legacy plaintext read-through is safe: `decryptPassphrase`/`decryptAddressPassphrase` return non-`encv1:`-prefixed values unchanged, so pre-existing docs keep working. ✔
- Wire format `encv1:base64url(iv|tag|ciphertext)` is self-describing, url-safe, version-prefixed, and parseable; length guard rejects bundles shorter than iv+tag. No double-wrap (`isEncrypted` short-circuit). ✔

Coverage/correctness: Encryption applied on source create and on PATCH (only when `body.address` provided); validation runs against the decrypted plaintext form; `toApi` decrypts before masking so `passphrase=***` masking still matches. `flow-generator` decrypts `src.address` immediately after DB fetch, and the `srt_uri` handed to Strom uses that decrypted value — decrypt coverage is complete for both the API and activation paths. Tests are thorough (key loading, round-trip, fresh-IV, tamper, legacy pass-through, address-level, masking, dev-fallback). `Co-Authored-By` trailer present.

OPERATIONAL RISK (explicit): `loadKey()` throws in production when `SRT_PASSPHRASE_KEY` is unset, and it is invoked lazily on the first source create/patch/read AND on flow activation. This fails closed — good for security — but merging + deploying before ops provisions `SRT_PASSPHRASE_KEY` will break source and production-activation flows in prod (and the key must remain stable, or already-encrypted docs become undecryptable). This is safe to MERGE into the codebase, but it is NOT safe to DEPLOY to production until ops has provisioned a stable 32-byte `SRT_PASSPHRASE_KEY`.

**MERGE-NOW: no** — gate the merge/rollout on ops confirming `SRT_PASSPHRASE_KEY` is provisioned (and backed up) in production first. Code itself is approved. Caller controls merge.

birme (commented):
APPROVE — Crypto reviewed carefully and is sound. AES-256-GCM via createCipheriv/createDecipheriv, fresh 12-byte random IV per encryption (randomBytes), 16-byte auth tag stored and setAuthTag() called on decrypt so tampering/truncation throws (verified by tests). Wire format encv1:base64url(iv|tag|ciphertext) is self-describing and length-checked (bundle.length < IV+TAG rejected). Key loaded from SRT_PASSPHRASE_KEY (base64 or hex, validated to exactly 32 bytes) — never hardcoded; fails closed in production, loud plaintext no-op in dev. No double-wrap (isEncrypted guard). Passphrase never logged. Route wiring correct: encrypt on POST/PATCH write, decrypt-on-read in toApi() before maskSrtPassphrase and in flow-generator before Strom; PATCH validates against decrypted address and only re-encrypts when address changes; legacy plaintext passes through. All referenced symbols (srtUrl, maskSrtPassphrase, SourceDoc.address) verified. Commit has Co-Authored-By trailer. One minor note (non-blocking): the passphrase-only encryption leaves streamid and host in cleartext by design, consistent with the stated scope of #160.

PR #175, opened 2026-08-27
fix(ws): make mixer tally writes durable on CouchDB 409 conflicts

### LINKED ISSUE #168: [BUG][HIGH] Mixer cut is broadcast to clients but never persisted when the tally write hits a CouchDB revision conflict
### Summary

Every mixer action writes the new tally back to the production document, but that write **swallows a
CouchDB 409 and continues**. When a conflict occurs the cut is still broadcast to all WebSocket
subscribers, so operators see the correct state, while the persisted `ProductionDoc.tally` silently
keeps the previous value. Nothing is logged. Any consumer reading tally from the database, or
following the `_changes` feed, misses that cut entirely and has no way to detect that it happened.

The same helper elsewhere in the codebase retries on 409, which is what makes this look like an
oversight rather than a decision.

### Where

`src/ws/controller.ts`, in each mixer-action branch. The `CUT` case at line 513:

```ts
const updated: ProductionDoc = { ...doc, tally: newTally, updatedAt: new Date().toISOString() };
await db.insert(updated).catch((err: any) => { if (err?.statusCode !== 409) throw err });
```

The same pattern appears at lines 545 (`TRANSITION`), 583 (`SET_PVW`), 658 and 684 (`TAKE`), and
816 to 850 (`MACRO_EXEC`). Verified at `8d89449`.

Compare `updateProductionDoc()` in `src/routes/productions.ts:38-60`, which re-reads and retries the
same conflict up to `MAX_DB_WRITE_RETRIES` times:

```ts
if (err instanceof Error && 'statusCode' in err && (err as { statusCode?: number }).statusCode === 409) {
  if (attempt < MAX_DB_WRITE_RETRIES - 1) continue;
}
```

So the codebase already has the correct pattern. The mixer-action paths, which are the highest
frequency writers to that document, are the ones that do not use it.

### Why this matters

Two independent reasons, and the second is the one that motivated this report.

**1. Persisted tally can disagree with what actually went to air.** `ProductionDoc.tally` is not
just a cache: it is the value used to restore state after a backend restart. In
`src/ws/controller.ts:1412-1419` a fresh connection seeds in-memory tally from the document when
memory is cold. If a conflict dropped the last write, a restart restores an **out-of-date PGM/PVW**
and the first client to reconnect is told the wrong source is on air.

**2. It removes the only durable record of a cut.** The tally in the production document, updated on
every action with an ISO 8601 `updatedAt`, is currently the only place in Open Live where a cut
leaves a persistent, timestamped trace. There is no event log and no `_changes` consumer in the
codebase. We are building a C2PA provenance sidecar that observes Open Live read-only and needs to
state, for each interval of a live programme, exactly which sources were on air and when. A cut that
is broadcast but not persisted is invisible to any consumer that was not connected at that instant,
and, worse, is indistinguishable from no cut at all.

We can design around a documented gap. We cannot design around a **silent** one, which is why this is
filed as a bug rather than a feature request.

### How a conflict arises

`doc` is fetched once at the top of `handleMessage` and every branch writes back a full document
built from it. Any concurrent writer to the same production invalidates the `_rev` in between. Known
concurrent writers include the activation polling loop (`src/routes/productions.ts`, which patches
`stromFlowId`, `status`, `whepOutputUrls` and endpoint fields while a production is activating), the
idle watchdog, and simply a second controller client, since the WebSocket fan-out explicitly supports
multiple subscribers per production.

The window is small, which is likely why this has not been noticed. It is widest exactly when the
system is busiest: during activation, and when more than one operator or automation client is
connected.

### Suggested direction

Route these writes through the existing `updateProductionDoc()` helper, or apply the same re-read and
retry loop inline. At minimum, log the dropped write at `warn` so the failure stops being silent.

### If this is deliberate

If swallowing the conflict is intentional, because the in-memory tally plus the broadcast is
considered the source of truth and the document is only a best-effort snapshot, that is a reasonable
design and we will build to it. In that case the useful fix is documentation: state that
`ProductionDoc.tally` is not authoritative and must not be relied on for history. That single
sentence would change our design.

### Reported by

Team building a C2PA provenance sidecar that observes Open Live read-only. Found by source review at
`8d89449`, not by a production incident. Happy to supply a failing test that forces the conflict if
that is useful.

birme:
## Summary
- Mixer actions broadcast the new tally but silently swallowed CouchDB 409 conflicts when persisting the production doc, so a cut/transition could be lost on reload (#168).
- Replaced the swallow pattern with a re-read-then-retry helper so mixer writes actually land under concurrent writes.

## Changes
- `src/ws/controller.ts`: Added `persistMixerMutation(productionId, action, mutate)`, mirroring `updateProductionDoc()`'s retry-on-409 loop. It re-reads the latest doc and re-applies the mutation against the fresh `_rev` on each retry, up to `MAX_DB_WRITE_RETRIES`. If every attempt still conflicts it logs at `warn` with `productionId` + action type instead of failing silently.
- Applied it to the CUT, TRANSITION, TAKE, SET_PVW and SELECT_PVW_PIP branches, plus the CUT / TRANSITION / TAKE / GRAPHIC_ON / GRAPHIC_OFF writes inside MACRO_EXEC.
- Removed the `.catch((err) => { if (err?.statusCode !== 409) throw err })` swallow and the bare inserts. Broadcast behaviour and external message shapes are unchanged.

## Test Plan
- [ ] Unit tests pass (see note below)
- [ ] Integration tests pass
- [ ] Manual verification of the 409 retry path:
  - Bring up a production, connect the studio WS, and issue a `CUT`.
  - To force a 409, race the write: e.g. add a temporary `await db.insert({ ...doc, updatedAt: new Date().toISOString() })` immediately before the mixer write (or fire two `CUT`s concurrently, or bump the doc's `_rev` out-of-band via `curl -X PUT .../open-live/<productionId>` between the handler's initial read and its insert). The first insert then hits `statusCode === 409`.
  - Before the fix: the 409 is swallowed, the DB tally is stale, and reloading the production shows the pre-cut PGM.
  - After the fix: `persistMixerMutation` re-reads the fresh `_rev`, re-applies `tally: newTally`, and the persisted doc matches the broadcast. Exhausting all retries logs `[controller] Failed to persist mixer mutation ...` at `warn` rather than dropping the change silently.

## Checklist
- [x] Code-quality checks passed (`npx tsc --noEmit` clean; see note on test suite)
- [x] No hardcoded SRT ports
- [ ] docs/repo-patterns.md updated if non-obvious issue encountered (n/a)

> Note: the vitest suite has 27 pre-existing failures on `main` in this environment (routes return 500 because CouchDB/Strom are not reachable). Results are identical with and without this change (27 failed / 22 passed), and the MACRO_EXEC tests pass in both — this PR introduces no new failures.

Closes #168

🤖 Generated with Claude Code

### DIFF
--- a/src/ws/controller.ts
+++ b/src/ws/controller.ts
@@ -18,6 +18,59 @@ function stromErrorMessage(err: unknown): string {
   return String(err);
 }
 
+// Mirror of MAX_DB_WRITE_RETRIES in routes/productions.ts — the number of
+// insert attempts a mixer write makes before giving up on a CouchDB 409.
+const MAX_DB_WRITE_RETRIES = 3;
+
+function isConflictError(err: unknown): boolean {
+  return (
+    err instanceof Error &&
+    'statusCode' in err &&
+    (err as { statusCode?: number }).statusCode === 409
+  );
+}
+
+/**
+ * Durably persist a mixer mutation to a ProductionDoc.
+ *
+ * CouchDB rejects an insert with 409 when the doc's `_rev` is stale (a
+ * concurrent write landed first). Previously these mixer writes swallowed the
+ * 409 and moved on, so the broadcast reflected a cut/transition that was never
+ * saved. Here we re-read the latest doc and re-apply the mutation against the
+ * fresh `_rev` before each retry, so the change actually lands. `mutate`
+ * receives the freshly read doc and returns the updated doc to insert;
+ * `updatedAt` is stamped automatically. If every attempt still conflicts the
+ * failure is logged at warn (with productionId + action) instead of silently
+ * dropped — it is no longer thrown, to keep the mixer broadcast path unchanged.
+ */
+async function persistMixerMutation(
+  productionId: string,
+  action: string,
+  mutate: (doc: ProductionDoc) => ProductionDoc,
+): Promise<void> {
+  const db = getDb();
+  for (let attempt = 0; attempt < MAX_DB_WRITE_RETRIES; attempt++) {
+    try {
+      const current = await db.get(productionId);
+      const updated: ProductionDoc = {
+        ...mutate(current),
+        updatedAt: new Date().toISOString(),
+      };
+      await db.insert(updated);
+      return;
+    } catch (err) {
+      // On a revision conflict, loop to re-read the latest _rev and re-apply.
+      if (isConflictError(err) && attempt < MAX_DB_WRITE_RETRIES - 1) continue;
+      console.warn(
+        `[controller] Failed to persist mixer mutation after ${attempt + 1} attempt(s)`,
+        { productionId, action },
+        err,
+      );
+      return;
+    }
+  }
+}
+
 type InboundMessage =
   | { type: 'CUT'; mixerInput: string; afvRampUpMs?: number; afvRampDownMs?: number }
   | { type: 'TRANSITION'; mixerInput: string; transitionType: string; durationMs?: number; afvRampUpMs?: number; afvRampDownMs?: number }
@@ -510,8 +563,7 @@ async function handleMessage(
         pvwPipByProduction.set(productionId, null);
         broadcast(productionId, { type: 'PIP_STATE', pgmPip: null, pvwPip: null, pips: pipConfigsByProduction.get(productionId) ?? [] });
       }
-      const updated: ProductionDoc = { ...doc, tally: newTally, updatedAt: new Date().toISOString() };
-      await db.insert(updated).catch((err: any) => { if (err?.statusCode !== 409) throw err });
+      await persistMixerMutation(productionId, 'CUT', (d) => ({ ...d, tally: newTally }));
       broadcast(productionId, { type: 'TALLY', ...newTally });
       await stromTransition(doc, fromPadCut, msg.mixerInput, 'cut');
       if (curPgmPipCut !== null && doc.stromFlowId && doc.mixerBlockId) {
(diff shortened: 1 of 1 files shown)


borisasadanin:
Reporter of #168. Two things: a regression the refactor introduces, and the test you asked us for.

## The retry widens error swallowing rather than narrowing it

The old line only swallowed a 409 and rethrew everything else:

```ts
await db.insert(updated).catch((err: any) => { if (err?.statusCode !== 409) throw err })
```

`persistMixerMutation` catches **every** error, and on any non-conflict error the `continue` guard is false, so it falls through to `console.warn` and `return`. A 500, a socket reset, a serialisation failure, anything that used to propagate is now logged and swallowed. `db.get(productionId)` is inside the same `try`, so a failed *read* is swallowed too, and that path did not previously exist at all.

Net: the fix makes conflicts durable and makes every other failure quieter than before. Suggested shape, keeping your deliberate decision not to throw on exhausted conflicts:

```ts
} catch (err) {
  if (isConflictError(err)) {
    if (attempt < MAX_DB_WRITE_RETRIES - 1) continue;
    console.warn(...);   // conflict exhausted: log, do not throw
    return;
  }
  throw err;             // everything else keeps its old behaviour
}
```

## The exhaustion path still diverges, which is worth stating in the issue rather than the code

After three conflicting attempts the cut is broadcast and not persisted, exactly as before, now with a log line. That is a fair reading of "or adding a `warn` log at minimum", and the restart-restores-the-wrong-source bug is genuinely much less likely. But it is mitigated, not eliminated, and nothing tells the client its cut did not stick.

We care about this more than most consumers: our sidecar treats the broadcast as the record of what happened, so broadcast-without-persist is a divergence between two sources of truth about the same cut. Not asking you to solve it here. Worth a sentence in #168 so it is not assumed closed.

Minor: three immediate retries with no backoff or jitter, against precisely the doc that is contended.

## The failing test

You asked on #168 for a test that forces the 409, and this PR has no test file. Our offer stands and this is now exactly the gap. We would write the shape we described there: assert the cut **is** broadcast, assert the tally **is** persisted after a forced conflict, and the cold-start assertion that the restored source is the one actually cut to.

Two things we need from you before we spend the effort:

1. **Where should it land?** We cannot push to a branch on your repository. So either you cherry-pick a patch we attach here, or we open a follow-up PR from a fork once this merges. Your call which is less annoying.
2. **How do you want the 409 forced?** A real conflicting write needs a CouchDB fixture and we do not know whether the suite has one. Stubbing `db.insert` to reject once with `{statusCode: 409}` is cheaper and would still fail against the old code. We would default to the stub for the retry assertion and the real fixture only if one already exists.

Also worth deciding: with the change above, a test asserting a non-409 error still propagates is one line and would pin the regression permanently.

birme:
🤖 daily-backlog-pr Phase 3 — **held, not merged**. CI is green, but this PR was authored under the same automation identity (`birme`) that this run operates as. Per the hard constraint *never self-review / self-merge*, this run cannot approve or merge its own PR — it needs review + merge by a different (human or independent) identity. Left in **In review**.

borisasadanin:
Pointer for anyone arriving here from #168: the review point on this PR about widened error swallowing was not addressed before merge, so it is now on `main`. Filed separately as #203 rather than re-litigating a merged PR, with the exact before-and-after, the five `MACRO_EXEC` sites that previously had no catch at all, and a concrete shape for the regression test you asked for on #168.

The retry itself does what #168 asked for, and #168 is correctly closed. #203 is only about the error classes that used to propagate.

PR #179, opened 2026-08-27
fix(ws): displace a PGM PiP when a macro cuts, transitions, or takes

wagenet:
**Builds on #182**, which adds the test harness. The diff contains both commits until that merges — only the second, `fix(ws): displace a PGM PiP when a macro cuts, transitions, or takes`, belongs to this PR.

## Problem

`CUT` and `TRANSITION` each do four things when a real source replaces a PiP that is on program: resolve `from_input` from `pgmBgByProduction`, move the PiP from PGM to PVW, broadcast `PIP_STATE`, and call `selectPreview` so the PiP lands back in Strom's preview.

The same actions run from a macro do none of it. `MACRO_EXEC` updates tally and calls `stromTransition` directly, so with a PiP on program it is never returned to preview, `pgmPipByProduction` and `pgmBgByProduction` keep their old values, and no `PIP_STATE` is emitted — leaving the server and every subscriber believing a PiP is on program long after it left.

## Fix

Mirrors the interactive logic into the three macro actions. The displacement block is inlined rather than extracted, so each macro path diffs directly against the handler it copies and the interactive paths stay untouched — happy to extract a shared helper instead.

## Impact

Verified against the Strom source rather than inferred. All of it is gated on `pgmPipByProduction` being set; **with no PiP on program, behaviour is unchanged.**

**The `selectPreview({ source: { pip: N } })` restore is the substance of this PR.** It sets `pvw_pip` in the block's overlay state, which `trigger_transition` reads as authoritative (`gst/pipeline/effects/take.rs`). Without it, preview holds the cut target that `stromTransition` selected rather than the displaced PiP, so the operator's next take airs the wrong source.

**The frame airing when the macro runs is unaffected** — the cut target comes from `stromTransition`'s own `selectPreview({ input: toIndex })`, which is untouched here.

**`fromPad` is included for consistency with the interactive handlers, not as a fix.** Strom ignores the request's `from_input`/`to_input` whenever overlay state exists, falling back to them only when it is missing entirely. By the same token the equivalent machinery in the interactive handlers is inert, and three comments in `controller.ts` (`:301`, `:592`, `:679`) describing `to_input` as setting Strom's `pgm_input` and averting the "sole program source" 400 are incorrect — that guard is in `mixer_ops.rs:80` and keys off overlay state. Left alone here to keep the diff to the macro paths; happy to correct them in a follow-up.

**A second, separate bug this does not fix.** A PiP sitting in *preview* is never promoted to program by a macro TAKE, and the failure is worse than stale state. `SELECT_PVW_PIP` leaves `tally.pvw` null, so a macro TAKE computes `{ pgm: null, pvw: <old pgm> }`, broadcasts it, and then passes that null to `stromTransition`, which early-returns on `!toMixerInput`. The server therefore tells every client the program bus is empty, Strom is never called and keeps airing the previous source, and `pgmPipByProduction` is never set. Fixing it means porting the interactive TAKE's PiP branch, which runs its own Strom sequence — larger than this PR and worth reviewing on its own. Happy to send it next.

## Conflicts with #156

Five hunks: three in `src/ws/controller.ts`, one per macro action, because both PRs touch the macro `TALLY` broadcast and the `stromTransition` call beside it; plus two in `src/__tests__/ws-macro-pip.test.ts`, because both add cases at the same insertion point.

All mechanical. The `controller.ts` ones resolve by keeping this PR's `fromPad` argument and restore block alongside #156's `pgmBg:` on the `TALLY` line; the test ones by keeping both sets of cases. No semantic overlap — #156 adds a broadcast field and changes no mixer behaviour. I will rebase whichever lands second.

## Testing

Ships with the two harness cases it makes pass — a macro CUT and a macro TRANSITION over a PiP on program. Both fail without the change; the no-PiP control from #182 keeps passing, so the untouched path stays pinned.

- `npx vitest run src/__tests__/ws-macro-pip.test.ts` — 3 passed.
- `npx tsc --noEmit` clean.
- `npx vitest run` — the suite's 27 pre-existing failures in `activation.test.ts` and `macros.test.ts` are unchanged and present on `main` too.
- **Not exercised against a live mixer.** Strom builds and runs locally, but neither of its sample flows carries a vision mixer block and its own suite has no PiP coverage, so Strom's response to `selectPreview({pip:N})` rests on reading `flows.rs` and `mixer_ops.rs` rather than on execution.

### DIFF
--- a/src/__tests__/ws-macro-pip.test.ts
+++ b/src/__tests__/ws-macro-pip.test.ts
@@ -1,0 +1,186 @@
+/**
+ * Harness and tests for PiP handling in macro-executed CUT, TRANSITION, and
+ * TAKE actions.
+ *
+ * The real `StromClient` runs against a throwaway HTTP server that records
+ * every request, so the assertions cover the URL, the verb, and the body the
+ * server actually puts on the wire — not a hand-written stand-in that could
+ * drift from the client without anything failing. CouchDB is mocked via
+ * vi.mock('../db/index.js'), as elsewhere in this suite.
+ *
+ * PiP state is established through real inbound messages (SELECT_PVW_PIP,
+ * TAKE) rather than by reaching into the module-level maps, so each case
+ * exercises the same state machine the server runs in production.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterAll } from 'vitest';
+import { createServer, type Server } from 'node:http';
+import type { AddressInfo } from 'node:net';
+
+// ---------------------------------------------------------------------------
+// Mock the CouchDB layer
+// ---------------------------------------------------------------------------
+
+const mockGet = vi.fn();
+const mockInsert = vi.fn().mockResolvedValue({ ok: true });
+
+vi.mock('../db/index.js', () => ({
+  getDb: () => ({ get: mockGet, insert: mockInsert, find: vi.fn().mockResolvedValue({ docs: [] }) }),
+  getSourcesDb: () => ({ get: mockGet, insert: mockInsert, find: vi.fn().mockResolvedValue({ docs: [] }) }),
+  connectDb: vi.fn().mockResolvedValue(undefined),
+  isDbReady: vi.fn().mockResolvedValue(true),
+}));
+
+vi.mock('../routes/productions.js', () => ({
+  updateProductionDoc: vi.fn().mockResolvedValue(undefined),
+}));
+
+vi.mock('../lib/strom-token.js', () => ({
+  getStromToken: vi.fn().mockResolvedValue(undefined),
+}));
+
+// ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


birme:
This PR is part of a multi-PR dependency stack (the WS macro / PiP fixes), in dependency order:
- Eyevinn/open-live#182 — WS-controller macro test harness (already MERGED)
- Eyevinn/open-live#179 — displace a PGM PiP when a macro cuts/transitions/takes
- Eyevinn/open-live#180 — take a preview PiP to program from a macro

Holding off on automated review/merge here — both open PRs are drafts and each still carries the other's commits until the one below it merges, so they need a human (or the relevant dev agent) to assess the stack as a whole, not a per-PR pass. Nothing further is needed from the author; this comment records the dependency on the PR itself rather than only in a run log.

PR #181, opened 2026-08-28
test(ws): cover PiP handling in macro CUT, TRANSITION, and TAKE

wagenet:
**Builds on #179 and #180.** The diff contains all three commits until those merge — only the third, `test(ws): cover PiP handling in macro CUT, TRANSITION, and TAKE`, belongs to this PR.

## Why

The macro PiP paths had no coverage, and neither does the WS controller as a whole — `activation.test.ts` and `macros.test.ts` both mock `../ws/controller.js` out entirely. #179 and #180 rested on reading the interactive handlers and mirroring them, which is exactly the kind of change that deserves execution behind it.

## What it does

Drives `handleMessage` with CouchDB and `StromClient` mocked, asserting the Strom call sequence and the broadcast payloads. PiP state is established through real inbound messages (`SELECT_PVW_PIP`, `TAKE`) rather than by reaching into the module-level maps, so each case exercises the same state machine the server runs in production.

Four cases:

| Case | Asserts |
| --- | --- |
| macro CUT over a PGM PiP | PiP moves to preview, `PIP_STATE` emitted, `selectPreview({pip:0})` called, `from_input` is the tracked background |
| macro TRANSITION over a PGM PiP | same displacement and restore |
| macro TAKE with a PiP in preview | Strom is called at all, PiP reaches program |
| macro CUT with no PiP | no `PIP_STATE`, no pip-addressed `selectPreview`, tally unchanged |

## It discriminates

A test that passes against broken code is worth nothing, so each case was run against the unfixed trees:

```
against main (neither fix)    3 fail, control passes
with #179 only                1 fails — the preview-PiP take
with both fixes               4 pass
```

The no-PiP control passes on all three, which is what makes it a control rather than a vacuous assertion. The single failure against #179 is precisely the bug #180 fixes.

## Production change

One export:

```ts
/** Exported for tests: drives one inbound message against a production. */
export async function handleMessage(
```

`handleMessage` was module-private and the plugin only exposes a websocket route, so there was no seam. The alternative — a fastify + websocket fixture — would drive the connect handler too and test far more than these paths. If you would rather have that, I am happy to switch; this felt like the smaller commitment to make on your behalf. Nothing else in `controller.ts` moves.

## Testing

- `npx tsc --noEmit` clean.
- `npx vitest run` — the new file passes 4/4. The suite's 27 pre-existing failures in `activation.test.ts` and `macros.test.ts` are unchanged (all 500s from the HTTP layer, present on `main` too).

**Still not covered:** anything requiring a real mixer. Strom builds and runs locally, but neither of its sample flows carries a vision mixer block and its own suite has no PiP coverage either, so `selectPreview({pip:N})` setting `pvw_pip` such that the next transition takes the PiP to program remains supported by reading `flows.rs` and `mixer_ops.rs` rather than by execution.

### DIFF
--- a/src/__tests__/ws-macro-pip.test.ts
+++ b/src/__tests__/ws-macro-pip.test.ts
@@ -1,0 +1,239 @@
+/**
+ * Tests for PiP handling in macro-executed CUT, TRANSITION, and TAKE actions.
+ *
+ * `handleMessage` is driven directly with CouchDB and StromClient mocked, so
+ * the assertions cover the Strom call sequence and the broadcast payloads
+ * without needing a mixer or a database.
+ *
+ * PiP state is established through real inbound messages (SELECT_PVW_PIP,
+ * TAKE) rather than by reaching into the module-level maps, so each case
+ * exercises the same state machine the server runs in production.
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// ---------------------------------------------------------------------------
+// Mock the CouchDB layer
+// ---------------------------------------------------------------------------
+
+const mockGet = vi.fn();
+const mockInsert = vi.fn().mockResolvedValue({ ok: true });
+
+vi.mock('../db/index.js', () => ({
+  getDb: () => ({ get: mockGet, insert: mockInsert, find: vi.fn().mockResolvedValue({ docs: [] }) }),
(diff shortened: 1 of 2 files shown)


wagenet:
Absorbed into the stack. The harness is now #182 at the root, and each fix ships with the cases it makes pass — #179 carries the two displacement cases, #180 the preview-PiP take. That way every PR is green on arrival instead of two behaviour changes landing ahead of their tests.

PR #183, opened 2026-08-28
docs: state the one-click CouchDB flow and the two undocumented Strom gates

alexbj75:
## What

The managed OSC flow ("Try it on OSC") already auto-provisions CouchDB and generates its own
admin password, but the README did not say so, and did not distinguish that flow from the
self-hosted `.env` setup further down. It also did not mention two gates that stop an operator
before they reach any further setup step:

- Shared Strom (the 69 EUR/month option) requires the Professional plan or above. The app shows
  this proactively when you open the environment, not as a failed submission on create.
- A self-hosted ("BYO") Strom instance has no plan requirement, but its URL must be publicly
  reachable; a plain LAN or loopback address is rejected. A mesh VPN (Tailscale, self-hosted
  Headscale) is one way to satisfy that without port-forwarding, and this path is token-metered
  on the free plan instead of plan-gated.

## Why

Source: an Open Live DX audit run 2026-08-27/28 (`osc-devrel-agent`), Headline Finding 1 and 2a/2b.
The finding traces to a live read of `Eyevinn/open-live-site`'s `POST /api/openlive/create` and to
the live app HTML/copy. Full audit:
https://github.com/Eyevinn (internal agent output, happy to share the source doc on request).

This is item 1 of a 4-item proposal (DR-88). Item 4 of that proposal (naming a specific external
switcher in the Studio wiki's Stream Type table) is deliberately not included here or anywhere:
a CEO decision logged 2026-08-27 bars naming a competing production switcher in interoperability
documentation, and that table is exactly that surface. The existing OBS/ffmpeg mention in that
table is unchanged; this PR does not touch the wiki.

## Not in this PR

- The community wiki's own Setup Guide page still describes the older four-manual-step CouchDB
  process, including a password-format warning that cannot occur on the flow this PR documents.
  That page needs the same correction, but GitHub wikis take no PR review and write access for
  this account is still pending a response from the maintainer (asked 2026-08-26). Flagging here
  so this PR is not read as having already fixed the wiki too.
- Any SRT/switcher-connection content. That is gated on confirming the SRT source address's
  caller/listener resolution behaviour with the team first (a separate, still-open item).

## Verification

Docs-only change, no code path affected. Rendered locally to confirm the added Markdown and the
`#environment-variables` anchor link resolve correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,24 @@ Visit **[openlive.apps.osaas.io](https://openlive.apps.osaas.io)** to spin up a 
 - 14-day free trial, free plan available
 - 15 EUR/month (self-hosted Strom) or 69 EUR/month (shared GPU in Frankfurt)
 
+The in-app **Create New Open Live** flow provisions CouchDB for you and generates its admin
+password automatically. You do not choose or handle that password yourself on this path.
+The `COUCHDB_URL` environment variable documented under [Environment variables](#environment-variables)
+below is only for a self-hosted deployment where you run CouchDB yourself; it does not apply
+to the managed OSC flow above.
+
+**Before you start, two things that are not obvious from the pricing line above:**
+
+- **Shared Strom (the 69 EUR/month option) requires the Professional plan or above.** If your
+  account is not eligible, the app shows this up front when you open the environment, not as a
+  failed submission after you try to create one.
+- **Your own ("BYO") Strom instance has no plan requirement, but its URL must be publicly
+  reachable.** A plain local-network or loopback address is rejected. A mesh VPN such as
+  [Tailscale](https://tailscale.com) or a self-hosted [Headscale](https://github.com/juanfont/headscale)
+  server is one way to give a locally hosted Strom instance a publicly routable address without
+  port-forwarding. On the free plan this path is metered by your token allowance rather than
+  gated by plan, so budget for that if you are testing rather than running on a paid plan.
+
 ## Features
 
 - **Vision mixing** — cuts, auto transitions, DSK layers, picture-in-picture, graphics overlays, and fade-to-black
--- a/README.md
+++ b/README.md
@@ -25,11 +25,15 @@ to the managed OSC flow above.
   account is not eligible, the app shows this up front when you open the environment, not as a
   failed submission after you try to create one.
 - **Your own ("BYO") Strom instance has no plan requirement, but its URL must be publicly
-  reachable.** A plain local-network or loopback address is rejected. A mesh VPN such as
-  [Tailscale](https://tailscale.com) or a self-hosted [Headscale](https://github.com/juanfont/headscale)
-  server is one way to give a locally hosted Strom instance a publicly routable address without
-  port-forwarding. On the free plan this path is metered by your token allowance rather than
-  gated by plan, so budget for that if you are testing rather than running on a paid plan.
+  reachable.** A plain local-network or loopback address is rejected, and so is any address
+  that is not reachable from OSC, including a plain Tailscale or other mesh-VPN address in the
+  `100.64.0.0/10` range even though it clears the initial check. To expose a locally hosted
+  Strom instance without port-forwarding, use a public ingress feature such as
+  [Tailscale Funnel](https://tailscale.com/kb/1223/funnel) (opt-in, off by default; also
+  implemented by self-hosted [Headscale](https://github.com/juanfont/headscale)), which gives
+  the instance a genuinely public `ts.net` hostname. On the free plan this path is metered by
+  your one-time token allowance rather than gated by plan, so budget for that if you are
+  testing rather than running on a paid plan.
 
 ## Features
 


alexbj75 (commented):
## Agent review gate: REQUEST CHANGES

Mandatory pre-merge agent review (`pr-review-toolkit:code-reviewer`) for an agent-opened PR. One blocking finding.

Every claim below was checked against `Eyevinn/open-live-site` source (the app that actually implements this flow) and against live vendor and pricing pages, not against the PR description.

### What checks out

- **Auto-provisioned CouchDB and generated admin password.** Confirmed in `open-live-site/src/server.ts`: `dbPassword` is generated with `crypto.randomBytes(...)` and passed straight to `createInstance(ctx, 'apache-couchdb', dbSat, { name: dbName, AdminPassword: dbPassword })`. The operator never picks or sees it. The README's correction is right.
- **"Professional plan or above" is the correct description of the shared-Strom gate.** The server accepts `PROFESSIONAL`, `PRO`, `ENTERPRISE`, `BUSINESS`. The live osaas.io plan ladder is Free, Personal 15 EUR, Professional 69 EUR, Business 199 EUR, Enterprise, so that set is exactly "Professional and above". Nothing understated.
- **"Shown up front, not as a failed submission" is accurate.** `src/public/index.html` fetches `/api/tenantplan` on load and toggles `#upgrade-overlay` on while hiding `#dashboard-content` before any create call. The 403 path exists as a backstop, but the proactive overlay is real.
- **BYO genuinely has no plan gate.** The BYO branch in `/api/openlive/create` skips the plan check entirely, and the BYO form is nested inside the upgrade overlay, so it is reachable on the free plan.
- **Placement.** Both gates sit in a bolded "Before you start" block immediately after the pricing line and above `## Features`, so a reader hits them before investing effort. This is the right position.
- **Links, anchor, style.** `tailscale.com`, `github.com/juanfont/headscale` and `openlive.apps.osaas.io` all return 200. `#environment-variables` resolves to the `## Environment variables` heading. No em dashes, no percent symbol, and no hardcoded service count in the added lines.
- **No contradictory leftovers.** The `+18/-0` shape is defensible here. The remaining CouchDB references in the README (Requirements, the env table, the OSC deployment section) all belong to the self-hosted path, which the new paragraph explicitly scopes. There was no incorrect manual instruction in this file to delete.

---

## Critical (blocking)

### 1. The Tailscale and Headscale advice describes something a mesh VPN does not do, and the reader will pass validation and then fail

`README.md`, lines 28 to 30 on the PR branch:

> A mesh VPN such as [Tailscale](https://tailscale.com) or a self-hosted [Headscale](https://github.com/juanfont/headscale) server is one way to give a locally hosted Strom instance a publicly routable address without port-forwarding.

A mesh VPN does not give a publicly routable address. That is the definition of what it does not do. Tailscale's own documentation states that tailnet devices get addresses from the RFC 6598 CGNAT range `100.64.0.0/10` and that "Tailscale IP addresses aren't exposed to the public internet". The feature that actually produces public reachability is **Tailscale Funnel**, which is a separate opt-in feature and is disabled by default. Headscale does implement Funnel, so naming Headscale is fine, but naming the mesh rather than Funnel is the error.

This is not a naming quibble, because of how it lands against the validator. The BYO check in `open-live-site/src/server.ts` rejects on hostname string matching only:

```
host === 'localhost' || /^127\./ || /^10\./ || /^172\.(1[6-9]|2\d|3[01])\./ ||
/^192\.168\./ || /^169\.254\./ || host === '::1' || host.endsWith('.local')
```

`100.64.0.0/10` is **not** in that list. So a reader who follows this README, installs plain Tailscale and pastes `http://100.x.y.z:7000` gets past the "must point to a publicly reachable host" guard, sees the create job start, and only then fails when the OSC-hosted server cannot reach a CGNAT address. A wrong instruction that clears the guard and fails opaquely downstream is worse than the gap this PR set out to close, and it is the same defect class: documentation describing a path the code does not support.

**Suggested fix**, docs-only:

> A plain local-network or loopback address is rejected, and so is any address that is not reachable from OSC, including a plain Tailscale or other mesh-VPN address in the `100.64.0.0/10` range even though it clears the initial check. To expose a locally hosted Strom instance without port-forwarding, use a public ingress feature such as Tailscale Funnel (opt-in, off by default; also implemented by self-hosted Headscale), which gives the instance a genuinely public `ts.net` hostname.

---

## Important (non-blocking)

### 2. "Budget for that" on the free plan omits that the free token grant is one-time

`README.md`, lines 32 to 33:

> On the free plan this path is metered by your token allowance rather than gated by plan, so budget for that if you are testing rather than running on a paid plan.

The metering claim is correct: `/api/tokens` reads `money.svc.prod.osaas.io/mytokencount`, and the BYO path is not plan-gated. But the live osaas.io pricing page lists the free plan as **100 tokens, one-time**, while the paid plans carry a recurring per-day allowance. "Budget for that" reads as a caution about a renewing allowance, which inverts the actual constraint for exactly the reader this sentence addresses: once the one-time grant is spent, the environment stops and nothing refills it.

Adding the word "one-time" to that sentence would resolve it. Not blocking on its own.

---

## Verdict

**REQUEST CHANGES** on finding 1. The correction this PR makes is real and the rest of it is accurate and well-placed, so this is a narrow fix to two sentences rather than a rethink.

Recorded as a COMMENT review, not an Approve: this PR was opened by `alexbj75` and GitHub does not accept a formal approval from the PR author's own account. Birme's team's merge remains the human gate.

alexbj75:
Fixed in 7a18c62, both applied.

**Blocking finding (mesh VPN vs. Funnel):** replaced the mesh-VPN-gives-a-public-address claim with your suggested wording, naming Tailscale Funnel (opt-in, off by default) as the actual public-ingress feature, and stating explicitly that a plain Tailscale/Headscale CGNAT address (`100.64.0.0/10`) clears the validator's blocklist but is not reachable from OSC. Headscale kept, since it implements Funnel too.

**Non-blocking (one-time token grant):** "budget for that" now reads "your one-time token allowance", so it no longer implies a renewing grant.

Not touching the vMix Stream Type table question, that stays withdrawn under the 2026-08-27 naming ruling and is out of scope for this PR either way.

birme:
Reviewed as part of daily-backlog-pr Phase 3.5 housekeeping. The content checks out: the repo-side claim is accurate — \`isPrivateHost\` in \`src/lib/url-validation.ts\` rejects RFC1918/loopback/link-local, and the README correctly notes that a \`100.64.0.0/10\` (CGNAT / plain Tailscale) address "clears the initial check" (there is indeed no CGNAT rule in that function) but is rejected for public-reachability platform-side. The plan-gating and token-metering statements are OSC-platform behavior outside this repo and read as consistent with the described flow. Docs-only, low risk.

I could not bring this PR up to date automatically: its branch predates the current \`.github/workflows/ci.yml\`, so \`update-branch\` returns 403 (\`refusing to ... without workflow scope\`) — this automation's token intentionally lacks \`workflow\` scope. A human maintainer needs to rebase this branch onto current \`main\` (which will queue CI) and merge. Flagging for human triage; not merging automatically.

birme (commented):
Reviewed as part of the daily housekeeping sweep (Phase 3.5). The prose is clear and the onboarding friction it documents is real and worth capturing — thank you.

**One thing blocks a merge from automation: I can't verify the factual claims from this repo.** All three assertions describe **OSC-platform / studio-frontend** behavior, not open-live backend behavior, and a code search of `Eyevinn/open-live` turns up no reference to any of them:

1. "Shared Strom (69 EUR/month) requires the Professional plan or above" — no plan-gating logic exists in this backend repo (it lives in the OSC platform / provisioning flow).
2. "A BYO Strom URL in the `100.64.0.0/10` range clears the initial check but is then rejected as unreachable" — the SSRF allow/deny logic in `src/lib/url-validation.ts` covers `10/8`, `172.16/12`, `192.168/16`, `127/8`, `169.254/16`; I could not confirm `100.64.0.0/10` (CGNAT) is in the rejected set, and the "clears the initial check then fails later" nuance isn't grounded in code here.
3. "On the free plan this path is metered by your one-time token allowance" — again a platform-billing behavior, not visible from this repo.

These may all be accurate from observed behavior, but I won't merge specific factual claims into the official README that I can't confirm. @svensson00 — could you confirm these three (Professional-plan gate for shared Strom, `100.64.0.0/10` rejection, free-plan token metering) are correct as written? Once confirmed, this is a clean docs merge. Flagging for maintainer verification rather than merging blind.

PR #184, opened 2026-08-31
fix(server): require API key auth for Swagger UI and OpenAPI spec routes

birme:
## Summary
- The API-key auth `onRequest` hook only fired for `/api/v1` and `/ws/` prefixes, so `/documentation` (Swagger UI) and its specs (`/documentation/json`, `/documentation/yaml`) were served unauthenticated even when `API_KEY` was set — exposing the full API blueprint (OWASP A01, CVSS 7.5). This extends the existing guard to also cover the `/documentation` prefix.

## Changes
- `src/server.ts`: added `!req.url.startsWith('/documentation')` to the auth-hook early-return so the Swagger UI and OpenAPI JSON/YAML spec routes require the same `Authorization: Bearer <API_KEY>` as the rest of the API. Reuses the existing guard/exemption logic — no new auth scheme. When `API_KEY` is unset, behaviour is unchanged (docs remain open, e.g. behind OSC's reverse-proxy auth wall).
- `src/__tests__/docs-auth.test.ts`: new focused vitest suite verifying unauthenticated requests to `/documentation`, `/documentation/json`, `/documentation/yaml` return 401, a wrong token returns 401, and a valid bearer token returns 200.

## Test Plan
- [x] Unit tests pass
- [x] Manual verification: unauthenticated request to Swagger UI / OpenAPI spec returns 401 when API_KEY set

## Checklist
- [x] No self-review

🤖 Generated with Claude Code

### DIFF
--- a/src/__tests__/docs-auth.test.ts
+++ b/src/__tests__/docs-auth.test.ts
@@ -1,0 +1,91 @@
+/**
+ * Tests that the Swagger UI and OpenAPI spec routes are protected by the same
+ * API-key auth guard as the rest of the API when API_KEY is configured.
+ *
+ * Regression test for the "Swagger UI / OpenAPI spec served without
+ * authentication" security issue (#81): the auth hook previously only fired for
+ * /api/v1 and /ws/ paths, leaving /documentation (+ /documentation/json,
+ * /documentation/yaml) exposed.
+ *
+ * CouchDB, the WS controller, and Strom are mocked — no real services required.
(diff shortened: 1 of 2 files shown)


birme:
Closing — this is our own earlier duplicate of #199 (same fix, same issue), which has since merged. Sorry for the churn — this is exactly the pipeline gap Phase 3.5 housekeeping is meant to catch going forward.

PR #190, opened 2026-08-31
fix(tests): stop cross-test state leakage in activation.test.ts

### LINKED ISSUE #188: activation.test.ts: 4 tests fail when run as a suite but pass in isolation (cross-test state leakage)
## Summary

Found while adding a CI workflow (see #<CI_PR>). `pnpm test` (vitest) currently fails 4 of 49
tests in `src/__tests__/activation.test.ts` when the file runs as a full suite, but every one
of them **passes in isolation** (`vitest run -t "<name>"`):

- `POST /api/v1/productions/:id/deactivate > clears whepEndpoint, stromFlowId, and mixerBlockId on deactivate`
- `POST /api/v1/productions/:id/deactivate > returns 200 even if production has no stromFlowId`
- `GET /api/v1/ice-servers > returns 502 if Strom is unreachable`
- `GET /api/v1/ice-servers > returns 502 if Strom returns a StromClientError`

All four fail the same way: an earlier test's **successful** response leaks into a later test
that mocked a rejection/error and expected a 502/error response, e.g.:

```
FAIL  GET /api/v1/ice-servers > returns 502 if Strom is unreachable
AssertionError: expected 200 to be 502
```

## What this isn't

Each `describe` block has a `beforeEach(() => vi.clearAllMocks())`, and every failing test
explicitly sets its own `mockResolvedValue`/`mockRejectedValue` on the relevant mock
(`mockStromSystemIceServers`, `mockGet`, etc.) before invoking the request — so this isn't the
classic "forgot to reset a shared mock" issue, and there are no `.mockOnce(...)` calls left
dangling in the file. `vi.clearAllMocks()` only clears call history, not implementations set via
plain `mockResolvedValue`/`mockRejectedValue` — but each test does set its own before the
request runs, so that alone doesn't explain the leak either.

## Working hypothesis

Since the failure only reproduces when tests run in the same process/file and always shows an
**earlier successful response** overriding a **later error expectation**, this looks like
shared, cached state somewhere in the request path — e.g. an in-memory cache for ICE servers or
production docs (in `src/lib/strom.js` or the route handler) that isn't scoped per `buildServer()`
call and isn't cleared between tests. Worth checking whether route handlers cache anything
module-level/keyed independent of the mocked client, and either clear that cache in `beforeEach`
or make the caching test-visible/injectable.

## Repro

```bash
pnpm install --frozen-lockfile
npx vitest run src/__tests__/activation.test.ts
```

## Why this matters now

This repo has no CI (see #4 on `Eyevinn/agent-team-open-live` and the security-team finding in
`open-live#101`/`open-live#7`) — a CI workflow is being added in a companion PR that runs
`pnpm test`. Until this is fixed, that workflow's test step will show 4 known failures on every
run; flagging here so that's not a surprise and so `tester`/`backend-developer` have a
documented starting point instead of rediscovering it from a red CI badge.

🤖 Filed while adding CI to this repo, on behalf of `Eyevinn/agent-team-open-live`.

birme:
## Summary
`npx vitest run src/__tests__/activation.test.ts` failed 4 of 49 tests only when run as a full suite. The failures traced to two independent test-isolation defects, both fixed here without weakening any assertions or disabling the stale-on-error cache feature.

Closes #188

## Changes
- **`src/routes/ice-servers.ts`** — export `resetIceServersCache()`. The stale-on-error cache (`cachedIceServers`) is module-scoped, so it correctly outlives a `buildServer()` rebuild in production but leaks between tests, which each build a fresh server yet share the module. An earlier test's successful ICE response was returned to a later test that mocked a rejection and expected a 502 (`expected 200 to be 502`). The reset hook lets tests clear it; the cache feature itself is untouched.
- **`src/__tests__/activation.test.ts`**
  - Call `resetIceServersCache()` in the ice-servers `beforeEach` so a cached success can't mask a later error expectation.
  - Complete the `../ws/controller.js` mock with `clearAudioState`, `clearPipState`, and `clearFxState`. The deactivate route imports these live-state cleanup helpers; the mock previously exported only `default`, so the handler threw on the missing export and returned 500 instead of 200.

## Test Plan
- [x] `npx vitest run src/__tests__/activation.test.ts` — 9/9 pass
- [x] `npx vitest run` (full suite) — 49/49 pass
- [x] `pnpm run typecheck` — clean
- [x] `pnpm run build` — clean

🤖 Generated with Claude Code

### DIFF
--- a/src/__tests__/activation.test.ts
+++ b/src/__tests__/activation.test.ts
@@ -6,6 +6,7 @@
 
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { buildServer } from '../server.js';
+import { resetIceServersCache } from '../routes/ice-servers.js';
 
 // ---------------------------------------------------------------------------
 // Mock CouchDB
@@ -29,6 +30,11 @@ vi.mock('../db/index.js', () => ({
 
 vi.mock('../ws/controller.js', () => ({
   default: async () => {},
+  // The deactivate route imports these live-state cleanup helpers; without them
+  // the mock is missing exports and the handler throws (→ 500) instead of 200.
+  clearAudioState: vi.fn(),
+  clearPipState: vi.fn(),
+  clearFxState: vi.fn(),
 }));
 
 // ---------------------------------------------------------------------------
@@ -239,6 +245,10 @@ describe('GET /api/v1/ice-servers', () => {
   beforeEach(() => {
     vi.clearAllMocks();
     mockFind.mockResolvedValue({ docs: [] });
+    // The ice-servers route keeps a module-level stale-on-error cache that
+    // survives across buildServer() instances. Clear it so a cached success
+    // from an earlier test can't be served in place of a mocked error (502).
+    resetIceServersCache();
   });
 
   it('returns 200 with iceServers array from Strom', async () => {
--- a/src/routes/ice-servers.ts
+++ b/src/routes/ice-servers.ts
@@ -20,6 +20,19 @@ import { config } from '../config.js';
 
 let cachedIceServers: IceServer[] | null = null
 
+/**
+ * Clear the module-level stale-on-error cache.
+ *
+ * The cache is module-scoped so it survives across `buildServer()` instances —
+ * that is correct in production (the last-good ICE config should outlive a
+ * server rebuild) but leaks state between tests, where each test builds a fresh
+ * server yet shares this module. Tests call this in `beforeEach` so a cached
+ * success from an earlier test can't mask a later error expectation.
+ */
+export function resetIceServersCache(): void {
+  cachedIceServers = null
+}
+
 const iceServersRoutes: FastifyPluginAsync = async (fastify) => {
   fastify.get('/api/v1/ice-servers', async (_req, reply) => {
     try {


birme (commented):
## PR Review — #190 fix(tests): stop cross-test state leakage in activation.test.ts

**Verdict: APPROVE**

### What Was Verified
- **No fabrication.** Both claimed fixes match real source:
  - `resetIceServersCache()` in `src/routes/ice-servers.ts` genuinely sets the real module-level `cachedIceServers` symbol (declared line 21) back to `null`. This is the exact symbol the stale-on-error handler reads (lines 46–48), so it correctly clears the leaking cache.
  - The deactivate route in `src/routes/productions.ts` really imports `clearPipState, clearAudioState, clearFxState` from `../ws/controller.js` (line 11) and calls all three (lines 553–555). These are genuinely exported from `src/ws/controller.ts` (lines 317/343/360). The mock completion is accurate, not invented. The mock also keeps the required `default` export (controller's real default at line 1655).
- **No weakening of tests.** No assertions deleted or loosened; the 6 `502` expectations remain. The only test changes are additive: mock exports + a `resetIceServersCache()` call in `beforeEach`.
- **Stale-on-error feature preserved for production.** The cache logic (lines 43–48) is untouched; reset is only invoked from the test `beforeEach`. Production behavior (last-good ICE config survives a `buildServer()` rebuild) is intact.
- **Scope is minimal:** only 2 files changed (`activation.test.ts`, `ice-servers.ts`).
- **Mandatory rules:** commit carries `Co-Authored-By:` trailer; branch is `fix/issue-188-activation-test-state-leak` (no direct push to main); no secrets added.
- **Checks green:** CI `build-and-test` = PASS. Ran `npx vitest run src/__tests__/activation.test.ts` locally → 9/9 passed.

### Critical Issues
None.

### Warnings
None.

Clean, root-cause fix. Recommend proceeding to the merge gate.

PR #217, opened 2026-09-09
Add pollable source providers and an open-weave source provider

KristalAlfred:
## Summary

Adds a source provider mechanism: a provider lists source candidates produced by
another system, and a poller reconciles them into CouchDB as ordinary source
documents. Assignment, activation and the studio UI need no knowledge of where a
source came from. The first provider lists sources from
[open-weave](https://github.com/Eyevinn/open-weave).

Providers are off by default. Nothing changes unless `SOURCE_PROVIDERS` is set.

## What is in here

**Provider interface** (`src/providers/types.ts`) — a provider is an id plus
`list(): Promise<ProviderSource[]>`. There is no push API; the registry polls. Every
candidate address is validated before it is stored: `srtUrl()` for `srt`/`efp`,
`httpUrlOnly()` for `whip`. The dispatch is exhaustive, so adding a stream type
to `ProviderSource` is a compile error until it has a validator.

**Registry and poller** (`src/providers/registry.ts`) — one reconciliation pass
per provider per tick:

| Candidate state | Action |
|---|---|
| Not in the DB | Create |
| Name, address, stream type, latency or status changed | Update |
| Unchanged | No write, so CouchDB is not touched every tick |
| Provider-owned doc no longer listed | Status set to `inactive`, doc kept |
| Provider or DB failure | Docs left as they are until the next tick |

A source that stops being listed is deactivated rather than deleted, because
deleting it would break the assignment of a production that still references it.
Invalid candidates are skipped with a reason and logged once per change, rather
than at the poll rate. Sync failures are logged on the transition into and out
of the failing state for the same reason.

**Read-only via the API** (`src/routes/sources.ts`) — provider-owned sources
carry `provider: { id, externalId, syncedAt }` and `readOnly: true` in responses.
`PATCH` and `DELETE` on them return `409`.

**The `weave` provider** (`src/providers/weave.ts`) — lists every enabled stream
on the northbound API and offers each placed, node-hosted output as an SRT source
with `?mode=caller` appended, which is the address open-live's
`builtin.mpegtssrt_input` block dials. Outputs weave dials out to, and streams
that are not yet placed, are skipped. The matching destination's declared SRT
latency is carried onto the source so the input block and the vision mixer's
`min_upstream_latency` agree with what weave configured; a value outside the
20–8000 ms the REST schema allows falls back to the default.

**Unknown provider ids fail startup** — `createSourceProviders` runs before the
server and DB come up, so a typo in `SOURCE_PROVIDERS` is a startup error rather
than a silently missing provider.

## The private-host flag

`srtUrl()` rejects private, loopback and link-local hosts so that an address in a
request body cannot make the pipeline dial internal services. A provider that
places media on a container or cluster network lists RFC1918 addresses for every
source — open-weave puts its SRT outputs on a node subnet — so the rule rejects
all of them.

`SOURCE_PROVIDER_ALLOW_PRIVATE_HOSTS=true` waives the rule for provider-listed
addresses only. Such an address is not attacker-supplied: it comes from the
system named in `SOURCE_PROVIDERS`, over the URL the operator configured. The
flag is off by default.

`srtUrl` and `httpUrlOnly` take the rule as an option rather than reading config,
so the REST routes keep it unconditionally and each caller states its own trust
assumption at the call site. `BLOCKED_HOSTNAMES` (`localhost`,
`metadata.google.internal`) still applies even when the flag is on.

## Configuration

| Variable | Purpose | Default |
|---|---|---|
| `SOURCE_PROVIDERS` | Comma-separated provider ids to poll | empty (disabled) |
| `SOURCE_PROVIDER_POLL_MS` | Poll interval in ms | `5000` |
| `SOURCE_PROVIDER_ALLOW_PRIVATE_HOSTS` | Accept provider-listed private hosts | `false` |
| `WEAVE_NORTHBOUND_URL` | open-weave northbound base URL | unset |
| `WEAVE_NORTHBOUND_TOKEN` | open-weave northbound bearer token | unset |

```bash
SOURCE_PROVIDERS=weave \
SOURCE_PROVIDER_ALLOW_PRIVATE_HOSTS=true \
WEAVE_NORTHBOUND_URL=http://localhost:29080 \
WEAVE_NORTHBOUND_TOKEN=<token> \
pnpm dev
```

README and `docs/openapi.yaml` are updated to match.

## Testing

`pnpm test` (196 tests, 19 files) and `pnpm typecheck` pass on this branch rebased
on `main` (`3293352`). New coverage:

- `src/__tests__/providers-registry.test.ts` — create, update, no-op, deactivate,
  duplicate and invalid candidates, `_rev` conflict retry, poll loop and stop.
- `src/__tests__/providers-weave.test.ts` — endpoint mapping, skipped outputs,
  latency handling.
- `src/__tests__/sources.test.ts` — `readOnly` in responses, `409` on `PATCH` and
  `DELETE`.
- Whip candidates: non-http scheme rejected, public https accepted, the
  private-host flag applied to whip, metadata hostname still blocked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,8 @@ Copy `.env.example` to `.env` and fill in the values:
 | `STROM_TOKEN` | OSC Personal Access Token for authenticating against an OSC-hosted Strom instance | _(empty — not needed for local Strom)_ |
 | `API_KEY` | Static API key protecting all `/api/v1` routes, the WebSocket controller, and the Swagger UI. **Required for any network-accessible deployment** (see below) | _(empty — routes unauthenticated)_ |
 | `LOG_LEVEL` | Fastify log level (`trace`, `debug`, `info`, `warn`, `error`) | `info` |
+| `SOURCE_PROVIDERS` | Comma-separated ids of [source providers](#source-providers) to poll. Unknown ids fail startup. | _(empty — disabled)_ |
+| `SOURCE_PROVIDER_POLL_MS` | Interval between source provider polls, in milliseconds | `5000` |
 
 > **Never commit `.env`** — it is gitignored. Use `.env.example` as the reference.
 
@@ -148,6 +150,12 @@ broadcasts — is documented separately in [`docs/controller-websocket.md`](docs
 Sources represent individual video/audio feeds. Each source has a `streamType` (`srt` or `whip`) and an `address` (SRT URI or WHIP endpoint URL).
 
 SRT passphrases are embedded in the source `address` and encrypted at rest before being stored in CouchDB. For rotating a passphrase or responding to a suspected compromise, see the operator runbook in [`docs/srt-passphrase-rotation.md`](docs/srt-passphrase-rotation.md).
+
+### Source providers
+
+A source provider is a small module under `src/providers/` that lists source candidates produced by another system. The server polls every provider named in `SOURCE_PROVIDERS` and materialises the candidates as ordinary sources, so assignment, activation and the studio UI need no knowledge of where a source came from.
+
+Provider-owned sources carry a `provider` object (`{ id, externalId, syncedAt }`) and `readOnly: true` in API responses. `PATCH` and `DELETE` on them return `409`. A source that the provider stops listing is marked `inactive` rather than deleted, so a production that still references it keeps its assignment.
 
 ### Template model
 
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ Copy `.env.example` to `.env` and fill in the values:
 | `LOG_LEVEL` | Fastify log level (`trace`, `debug`, `info`, `warn`, `error`) | `info` |
 | `SOURCE_PROVIDERS` | Comma-separated ids of [source providers](#source-providers) to poll. Unknown ids fail startup. | _(empty — disabled)_ |
 | `SOURCE_PROVIDER_POLL_MS` | Interval between source provider polls, in milliseconds | `5000` |
+| `WEAVE_NORTHBOUND_URL` | Base URL of the open-weave northbound API (e.g. `http://localhost:29080`). Required when `weave` is in `SOURCE_PROVIDERS`. | _(unset)_ |
+| `WEAVE_NORTHBOUND_TOKEN` | Bearer token for the open-weave northbound API. Required when `weave` is in `SOURCE_PROVIDERS`. | _(unset)_ |
 
 > **Never commit `.env`** — it is gitignored. Use `.env.example` as the reference.
 
@@ -156,6 +158,21 @@ SRT passphrases are embedded in the source `address` and encrypted at rest befor
 A source provider is a small module under `src/providers/` that lists source candidates produced by another system. The server polls every provider named in `SOURCE_PROVIDERS` and materialises the candidates as ordinary sources, so assignment, activation and the studio UI need no knowledge of where a source came from.
 
 Provider-owned sources carry a `provider` object (`{ id, externalId, syncedAt }`) and `readOnly: true` in API responses. `PATCH` and `DELETE` on them return `409`. A source that the provider stops listing is marked `inactive` rather than deleted, so a production that still references it keeps its assignment.
+
+Available providers:
+
+| Id | System | Configuration |
+|---|---|---|
+| `weave` | [open-weave](https://github.com/Eyevinn/open-weave) | `WEAVE_NORTHBOUND_URL`, `WEAVE_NORTHBOUND_TOKEN` |
+
+The `weave` provider lists every enabled stream on the northbound API and offers each placed, node-hosted output as an SRT source with `?mode=caller` appended, which is the address open-live's `builtin.mpegtssrt_input` block dials. Outputs that weave dials out to (remote destinations) and streams that are not yet placed are skipped. The matching destination's declared SRT latency is carried onto the source, so the input block and the vision mixer's `min_upstream_latency` agree with what weave configured; a value outside the 20–8000 ms the REST schema allows is ignored in favour of the default.
+
+```bash
+SOURCE_PROVIDERS=weave \
+WEAVE_NORTHBOUND_URL=http://localhost:29080 \
+WEAVE_NORTHBOUND_TOKEN=<token> \
+pnpm dev
+```
 
 ### Template model
 
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ Copy `.env.example` to `.env` and fill in the values:
 | `SOURCE_PROVIDERS` | Comma-separated ids of [source providers](#source-providers) to poll. Unknown ids fail startup. | _(empty — disabled)_ |
 | `SOURCE_PROVIDER_POLL_MS` | Interval between source provider polls, in milliseconds | `5000` |
 | `SOURCE_PROVIDER_ALLOW_PRIVATE_HOSTS` | Set to `true` to accept provider-listed SRT addresses on private, loopback or link-local hosts. See [source providers](#source-providers). | `false` |
+| `SOURCE_PROVIDER_ALLOWED_HOSTS` | Comma-separated hostnames, IPs and CIDR blocks a provider may name (e.g. `10.42.0.0/16`). When set, any other host is skipped and `SOURCE_PROVIDER_ALLOW_PRIVATE_HOSTS` is not consulted. See [source providers](#source-providers). | _(empty — not enforced)_ |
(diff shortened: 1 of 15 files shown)


birme:
Reviewed as part of daily-backlog-pr Phase 3.5 housekeeping. This is well-engineered — off-by-default provider sync, bounded reconciliation, tests, and the \`srtUrl(allowPrivateHosts)\` relaxation is opt-in (\`SOURCE_PROVIDER_ALLOW_PRIVATE_HOSTS\`, off by default), correctly scoped to provider-sourced (non-request-body) addresses, and explicitly leaves the REST routes' SSRF rule intact. No backdoor, no hardcoded secrets.

However, this is not a mechanical merge candidate: it introduces a new subsystem (provider registry/poller), a new external dependency (open-weave northbound API), a new config surface, and a security-relevant opt-in that relaxes the SSRF guard under an env flag. Per the epic-workflow, that combination is an architectural decision that warrants an architect spec + ADR and human sign-off before merge. Flagging for human/architect triage rather than auto-merging.

birme (changes requested):
**Verdict: REQUEST CHANGES (minor)** — reviewed as part of the daily housekeeping sweep (Phase 3.5). No blocking security issues; the feature is well-built and CI (`build-and-test`) is green.

**Security: sound.** The `allowPrivateHosts` bypass in `src/lib/url-validation.ts` defaults off and is only passed by the provider registry for provider-sourced addresses (not request-supplied). All REST call sites still call `srtUrl()` with no options, so the request-path SSRF rule is unchanged. No hardcoded secrets (Weave token via `WEAVE_NORTHBOUND_TOKEN`/`requireEnv`, sent as a Bearer header, with a test asserting it isn't leaked in errors). No command injection — fetch-only, stream names `encodeURIComponent`-escaped. Provider addresses encrypted at rest via the existing `encryptAddressPassphrase` pattern.

**Please address before merge:**
1. **Validation gap on `whip` candidates.** `validateCandidate` only runs `srtUrl()` for `srt`/`efp` stream types — a `whip` candidate address (allowed by `ProviderSource.streamType`) is stored unvalidated. It's latent today since the shipped `weave` provider only emits `srt`, but it's a real gap worth closing now rather than shipping a validation hole for the next provider to trip over.
2. **Rebase required.** Branch is `BEHIND` main — please rebase onto current `main` before this can merge.

Code quality is strong overall: fits the Fastify/TS/pnpm/vitest conventions, ESM `.js` imports, config centralised in `src/config.ts`, sensible reconciliation logic, ~655 lines of tests with DB/fetch mocked, and OpenAPI + README updated. No scope creep. Once the whip-validation gap is closed and the branch is rebased green, this is a clean merge.

KristalAlfred:
@birme Both points addressed.

**whip validation.** `validateCandidate` now covers every stream type. srt and efp still go through `srtUrl()`. whip goes through `httpUrlOnly()`, so the address has to be http or https on a non-private host. The dispatch is a switch with a `never` default, so adding a type to `ProviderSource` is a compile error until it has a validator.

`SOURCE_PROVIDER_ALLOW_PRIVATE_HOSTS` applies to whip the same way it applies to srt. I moved the private-host rule into an option on `httpUrlOnly` to do that. `BLOCKED_HOSTNAMES` (`localhost`, `metadata.google.internal`) still applies even with the flag on, so the metadata endpoint stays blocked either way. `httpUrlOnly`'s default behaviour is unchanged and every existing caller passes no options.

Four tests added in `providers-registry.test.ts`: bad whip scheme rejected, public https accepted, flag applied to whip, metadata hostname still blocked.

Two things I ran into while doing it:

- The REST path has the same gap. `SourceInput`, `SourcePatch` and the PATCH handler in `routes/sources.ts` also skip validation when `streamType` is `whip`. I left it alone, because tightening it would reject addresses the API accepts today, and that is a behaviour change beyond this PR. Happy to do it here or as a follow-up issue, your call.
- `flow-generator.ts` builds the whip input from `endpoint_id` and never reads `source.address`, so nothing dialled the unvalidated address. The gap was real but not reachable yet.

**Rebase.** Done, but I force-pushed over your merge commit `0785c0c`. It merged main at `d4400e5`, which was already behind by the time it landed. The branch now sits on `3293352` and includes #225, #226 and #227. Nothing from your merge was lost, it was a clean auto-merge with no conflict resolution in the provider files, but sorry for overwriting your commit.

`pnpm test` (196 tests, 19 files) and `pnpm typecheck` pass.

birme (changes requested):
Thanks for this — reviewed as part of the team's housekeeping sweep of open PRs. The defensive design is genuinely good: candidates are re-validated in the registry before touching CouchDB, latency is clamped to the same bounds as the REST schema, and the whole feature is inert unless `SOURCE_PROVIDERS` is set. Test coverage is strong.

**Blocker (security):** the SSRF gate is bypassable via the weave-supplied URL when `SOURCE_PROVIDER_ALLOW_PRIVATE_HOSTS=true` (the documented config for open-weave on a node subnet). In `weave.ts`, the source address is built from the northbound-returned `output.url` and only has `?mode=caller` appended — it is not re-parsed against the northbound origin. `output.url` is upstream-controlled data, not operator config, so a compromised/misbehaving northbound could make open-live dial any private/loopback/link-local SRT host. Please pin allowed hosts to the northbound host/subnet, or validate `output.host`/`output.port` rather than trusting the free-form `url`. Also consider running `WEAVE_NORTHBOUND_URL` through `httpUrlOnly` at construction (an `http://` scheme + Bearer token can otherwise traverse plaintext to a private host).

**Also:** the branch is `CONFLICTING` against current `main` — please rebase `sp-upstream-providers` onto `main` alongside the SSRF fix, and I'll re-review to merge.

KristalAlfred:
@birme Rebased onto main and added an allow-list. One part of the suggestion I want to push back on.

**The allow-list.** `SOURCE_PROVIDER_ALLOWED_HOSTS` takes hostnames, IPs and CIDR blocks:

```
SOURCE_PROVIDER_ALLOWED_HOSTS=10.42.0.0/16,weave-1.internal
```

When set, a candidate whose host is not on the list is skipped, and `SOURCE_PROVIDER_ALLOW_PRIVATE_HOSTS` is not consulted. It bounds the address regardless of what weave returns. It also catches something the private-host rule never looked at: `srt://attacker.example.com:9000` passes `srtUrl()` today, because that rule only rejects private IP literals and DNS names are out of scope for it. A malformed entry throws at startup so a typo doesn't silently match nothing. 19 new tests across `url-validation.test.ts`, `providers-registry.test.ts` and `providers-weave.test.ts`.

**Where I didn't follow the suggestion.** Validating `output.host` and `output.port` instead of `output.url` doesn't move the trust boundary. Both fields come out of the same northbound response. If the northbound is returning addresses it shouldn't, it controls `host` and `port` exactly as much as it controls `url`, so re-parsing swaps which field we trust rather than adding a check the attacker sits outside of.

Pinning to the northbound host or subnet has a different problem: weave puts SRT outputs on the node subnet, not on the northbound host, so that rule would reject the addresses the provider exists to carry. That's the topology the `SOURCE_PROVIDER_ALLOW_PRIVATE_HOSTS` paragraph in the README describes.

The allow-list is the version that works, because the operator supplies it. It's the one input to candidate validation that weave doesn't produce.

**On the severity.** I think the flag is doing less work than the review gives it credit for. With it off, `srtUrl()` still accepts any public host. So a northbound that has gone bad can already point Strom at `srt://attacker.example/...` and feed arbitrary media into a live production, no flag involved. What the flag adds on top is internal hosts, and that direction is blind: `weave.ts` hardcodes `status: 'active'`, nothing writes the dial result back into the source doc, and `SRT_URL_RE` locks the scheme, so it's a UDP connect with no response path. Worth closing, and now it is. But the bigger case was media injection from a public host, and the allow-list is the only one of the two that touches it.

**`WEAVE_NORTHBOUND_URL`.** Added a check, but scheme only, not `httpUrlOnly`. `httpUrlOnly` rejects `localhost` via `BLOCKED_HOSTNAMES`, and `http://localhost:29080` is the documented setup for a node-local weave (it's in the README example and in `providers-weave.test.ts`). The host there is the operator's choice, so I only assert it's http or https.

`pnpm test` (228 tests, 21 files) and `pnpm typecheck` pass.

birme:
🤖 Housekeeping (Phase 3.5): leaving this **with changes requested** and flagging for human/architect triage. Two blockers: (1) the branch is now `CONFLICTING` against `main` and needs a rebase before it can merge; (2) it already has unaddressed `CHANGES_REQUESTED` review feedback. Beyond that, this introduces a whole new source-provider subsystem (`src/providers/*`, poller, DB reconciliation, API surface) — that is epic-scoped and should go through an architect spec per `epic-workflow.md` rather than an autonomous orphan merge. Not closing it; it needs a human owner.

PR #229, opened 2026-09-10
feat(port-lease): lease SRT listener ports from the shared Strom

Lufftre:
## Why

Several Open Live instances share one Strom, and every hostless SRT listener source or output (`srt://:PORT?mode=listener`) binds a port on it. Nothing kept two instances, or the gateways behind them, from picking the same port; the second pipeline then failed at activation.

## What

Open Live now leases a block of consecutive ports and only accepts listener addresses inside it. The lease requests go to `STROM_URL` under `/api/port-leases`. On a shared Strom a proxy routes that path to a port broker service (strom-port-broker) that owns the pool; everything else passes through to Strom. A self-hosted Strom with no proxy answers 404, and Open Live applies no restriction there. Strom itself is unchanged.

- `src/services/port-lease.ts`: acquires at boot under a stable client id (`STROM_PORT_LEASE_CLIENT_ID`, else the `PUBLIC_BASE_URL` hostname, else `open-live-<hostname>`), renews every minute, re-acquires under the same id if the lease vanished, releases on `SIGTERM`/`SIGINT`. Any 404 on the create call means no broker: logged once, no enforcement, re-probed every 10 minutes.
- `GET /api/v1/server-info` publishes `srtPortRange: {first,last} | null` and `srtPortLease: leased | pending | unsupported | disabled`, so gateways (Eyevinn/open-live-ingest#1) pick ports from the leased block.
- Inside the block, Open Live assigns the ports (`src/services/listener-ports.ts`). A listener address with port 0 (`srt://:0?mode=listener`) asks for the lowest port no other source or output holds; the stored address and the response carry the assigned port. Gateways register this way, so several can feed one instance. An explicit port must be inside the block and unique, or the request fails with 422 naming the range or 409 naming the holder. A PATCH with port 0 keeps the document's current port when still valid. After writing an assigned port the routes look again and move to another if a concurrent request took the same one.
- `POST`/`PATCH /api/v1/sources` and `/api/v1/outputs` answer 503 while the lease is still pending. Caller addresses, WHIP/WHEP/HTML, and renames that leave the address alone are not checked, so existing documents keep working until their address is edited.
- Env: `STROM_PORT_LEASE_SIZE` (default 10, covering sources and outputs), `STROM_PORT_LEASE_CLIENT_ID`, `STROM_PORT_LEASE_DISABLED`.
- Docs: `docs/port-lease.md`, README env table, `.env.example`, `docs/openapi.yaml`.

Judgement calls worth a look: a hostless address with no `mode` parameter is treated as a listener, since it can only bind; on a non-404 renew failure the last known range is kept rather than dropping to `pending`, so sources keep validating against it while the broker is briefly unreachable.

Deploy note: once an instance holds a lease, gateways still on the old default range 47110-47129 get 422 unless that range falls inside the block. Upgrade the venue gateways together with this, or set `STROM_PORT_LEASE_DISABLED=true` until they are.

Studio follow-up (not in this PR): the outputs panel defaults to `srt://:43524?mode=listener`, which a shared Strom will now reject; Studio should propose a port from `srtPortRange`.

## Tests

- `pnpm typecheck`: clean
- `pnpm test`: 14 files, 190 passed. New: `src/__tests__/port-lease.test.ts` (service: acquire → renew → 404 → re-acquire, 409 retry, unsupported detection from any 404 body and back-off, disabled, stop/release), `src/__tests__/listener-ports.test.ts` (assignment, keep, uniqueness, exhaustion, no-range cases) and `src/__tests__/port-lease-routes.test.ts` (server-info in all four states; sources and outputs POST/PATCH: assignment, 422, 409, 503, pass-through, race retry, rename does not re-check).
- Smoke tested end to end on one machine (see comments below), and against the real broker deployed in front of the shared dev Strom: two gateways started 12 ms apart got 47110 and 47111, a crashed gateway got its port back on restart, and an activated production went on air.
- No lint script exists in the repo.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -43,7 +43,8 @@ API_KEY=change-me-before-use
 # Client id sent to Strom. Defaults to the hostname of PUBLIC_BASE_URL, or
 # open-live-<hostname> when PUBLIC_BASE_URL is unset. Keep it stable across restarts.
 # STROM_PORT_LEASE_CLIENT_ID=
-# Set to true to skip leasing entirely (single-tenant Strom, or a Strom without the API).
+# Set to true to skip leasing entirely. Not needed for a Strom with no port broker in
+# front of it: the 404 is detected and no restriction is applied.
 # STROM_PORT_LEASE_DISABLED=false
 
 # AES-256 key (32 bytes, base64 or hex) used to encrypt SRT source passphrases at
--- a/.env.example
+++ b/.env.example
@@ -36,6 +36,16 @@ API_KEY=change-me-before-use
 # If not set, the URL is derived from the incoming request (safe only with trustProxy configured).
 # PUBLIC_BASE_URL=https://your-open-live-instance.example.com
 
+# SRT port lease on a shared Strom instance. At startup Open Live reserves a range
+# of consecutive SRT listener ports from Strom and only accepts listener sources
+# (srt://:PORT?mode=listener) whose port falls inside that range.
+# STROM_PORT_LEASE_SIZE=20
+# Client id sent to Strom. Defaults to the hostname of PUBLIC_BASE_URL, or
+# open-live-<hostname> when PUBLIC_BASE_URL is unset. Keep it stable across restarts.
+# STROM_PORT_LEASE_CLIENT_ID=
+# Set to true to skip leasing entirely (single-tenant Strom, or a Strom without the API).
+# STROM_PORT_LEASE_DISABLED=false
+
 # AES-256 key (32 bytes, base64 or hex) used to encrypt SRT source passphrases at
 # rest in CouchDB. REQUIRED in production (NODE_ENV=production) — the server fails
 # to start without it. Left unset in local dev, passphrases are stored in plaintext
--- a/README.md
+++ b/README.md
@@ -73,6 +73,9 @@ Copy `.env.example` to `.env` and fill in the values:
 | `STROM_TOKEN` | OSC Personal Access Token for authenticating against an OSC-hosted Strom instance | _(empty — not needed for local Strom)_ |
 | `API_KEY` | Static API key protecting all `/api/v1` routes, the WebSocket controller, and the Swagger UI. **Required for any network-accessible deployment** (see below) | _(empty — routes unauthenticated)_ |
 | `LOG_LEVEL` | Fastify log level (`trace`, `debug`, `info`, `warn`, `error`) | `info` |
+| `STROM_PORT_LEASE_SIZE` | Number of SRT listener ports to lease from a shared Strom — see [`docs/port-lease.md`](docs/port-lease.md) | `20` |
+| `STROM_PORT_LEASE_CLIENT_ID` | Stable lease client id sent to Strom | hostname of `PUBLIC_BASE_URL`, else `open-live-<hostname>` |
+| `STROM_PORT_LEASE_DISABLED` | Set to `true` to turn off port leasing | `false` |
 
 > **Never commit `.env`** — it is gitignored. Use `.env.example` as the reference.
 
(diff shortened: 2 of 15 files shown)


Lufftre:
End-to-end smoke test on one machine, all three branches together (Strom `feat/port-leases`, Open Live `feat/strom-port-lease`, gateway `feat/port-range-from-open-live`), Strom on 8080 with `STROM_PORT_LEASE_RANGE=47100-47139`, Open Live against a local CouchDB, gateway `up --test` in caller mode with no `port_range`:

- Open Live acquired 47100-47109 at boot and published it on `server-info`; the gateway printed "SRT ports 47100-47109 from Open Live", registered `srt://:47100?mode=listener`, and the flow reached Playing
- a second tenant got 47110-47119; a flow created directly on Strom listening on 47125 made a 20-port request fail with 409 and a 14-port request land on 47126-47139; a 5 s TTL lease disappeared after expiry
- Open Live rejected a source and an output on port 43524 with 422 naming the range, accepted an in-range output and a caller-form source, and rejected a PATCH to an out-of-range address
- the lease renewal moved the expiry; leases survived a Strom restart with the same ranges
- SIGTERM on Open Live released its lease and a restart re-acquired 47100-47109; the gateway warned and kept its feed running meanwhile
- Ctrl-C on the gateway removed its flow and its Open Live source

birme (changes requested):
Thanks for this — reviewed as part of the team's housekeeping sweep of open PRs. The code itself is in good shape: every Strom/route contract it depends on actually exists, enforcement is fail-open-then-tighten (unsupported/disabled pass through, pending → 503, leased enforces range), and it's well covered by tests. No security or correctness blockers found (no secrets, no injection, the `parseListenerPort` regex is anchored and bounded, the lease check runs on plaintext address before encryption).

**Blocker:** the branch is currently `CONFLICTING` against `main` — `main` has advanced past your branch point (likely conflicts in `status.ts` / `sources.ts` / `outputs.ts`, which changed on both sides). Please rebase `feat/strom-port-lease` onto current `main` and resolve, and I'll re-review to merge.

Minor, non-blocking:
- `isUnsupportedError` and `isNotFound` are byte-identical (both `status === 404`) — a create-time vs renew-time 404 are semantically different; fine today but a latent trap.
- `evaluateListenerPort` treats a hostless SRT address with no `mode` param as a listener — reasonable but slightly aggressive; worth a maintainer sign-off.

Lufftre:
Thanks for the review. Addressed:

- **Rebased onto current `main`** (759e006 and later). The only overlap was in `src/config.ts`, where `main` added `trustedHosts` next to the lease settings; both are kept. Typecheck is clean and all 241 tests pass. GitHub now reports the branch mergeable.
- **`isUnsupportedError` / `isNotFound`**: collapsed into one `isNotFound` in 54a70a4. The two meanings now live as comments at the call sites: on create, a 404 means nothing serves the lease routes (a Strom with no port broker proxied in front of it), so leasing is unsupported here; on renew, the routes work but this lease is gone, so it is re-acquired. The reason they converged is deliberate: the lease routes are served by a separate broker behind a proxy (`Eyevinn/strom-port-broker`), so the 404 body on create is whatever the responder puts in it and only the status can be trusted.
- **Hostless address without `mode` treated as a listener**: kept, and now in `listenerPortRequest` in `src/services/listener-ports.ts`. Two reasons. Open Live's own `srtUrl` validator already documents the hostless form as "binds all interfaces", so it is this repo's convention that `srt://:PORT` means listen. And GStreamer's SRT elements do the same: `srtsink uri=srt://:8888` is the documented way to make it listen, while a host makes it call, so a hostless address without a mode does bind the port on Strom whatever we call it here. Treating it as anything else would let a colliding port through. An explicit `mode=caller` or `mode=rendezvous` on a hostless address is still left alone.

Since your review the branch also gained port assignment inside the leased range (`srt://:0?mode=listener` asks Open Live for a free port; explicit ports must be unique, 409 naming the holder), which the PR description now covers. It was live-tested against the shared dev Strom with two gateways registering 12 ms apart.

Issue #39, opened 2026-06-03
[SECURITY][MEDIUM] Fastify trustProxy not configured — X-Forwarded-Host from any client hijacks stored WHIP endpoint URLs


LucasMaupin:
## Summary
Fastify's `trustProxy` option is not set, but the production activation route manually reads `X-Forwarded-Proto` and `X-Forwarded-Host` from `req.headers` without verifying the request came from a trusted proxy. An unauthenticated attacker can supply `X-Forwarded-Host: attacker.example` in an activation request, causing the stored WHIP callback URL in CouchDB to point to `https://attacker.example/api/v1/...`. Browser clients then send WHIP ICE candidates (containing WebRTC connection information) to the attacker-controlled server. This is related to but distinct from issue #16.

## CVSS Score
5.8 (Medium) — CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:C/C:L/I:L/A:N

## Category
OWASP A05 — Security Misconfiguration / Header Injection

## Affected File(s)
- `src/routes/productions.ts:473-479`
- `src/server.ts` (Fastify initialization — trustProxy not set)

## Steps to Reproduce
```bash
curl -X POST http://localhost:8080/api/v1/productions/<id>/activate \
  -H 'X-Forwarded-Host: attacker.example' \
  -H 'X-Forwarded-Proto: https'
# Production activates with storedWhipEndpoint = https://attacker.example/api/v1/...
# Any browser client fetching the production will send WHIP ICE candidates to attacker.example
```

## Remediation
Choose one of:

**Option A** — Configure `trustProxy` and use Fastify's parsed values:
```typescript
// src/server.ts
const fastify = Fastify({
  trustProxy: '10.0.0.0/8',  // Only trust proxy headers from known proxy CIDR
});
// In productions.ts, use req.protocol and req.hostname (Fastify sets these correctly)
const publicBaseUrl = `${req.protocol}://${req.hostname}`;
```

**Option B** — Derive publicBaseUrl from an environment variable (safest):
```typescript
const publicBaseUrl = config.publicBaseUrl  // e.g., https://live.example.com
  ?? `${req.protocol}://${req.hostname}`;
```

## References
- [Fastify trustProxy documentation](https://fastify.dev/docs/latest/Reference/Server/#trustproxy)
- [OWASP HTTP Host Header Attacks](https://portswigger.net/web-security/host-header)
- Related: #16 (X-Forwarded-Host activation polling)

[...]

PR #232, opened 2026-09-11
fix(productions): validate WHIP callback host / configure trustProxy

### LINKED ISSUE #50: [SECURITY][MEDIUM] X-Forwarded-Proto read raw in WHIP callback URL — trustProxy not configured
## Summary
When \`PUBLIC_BASE_URL\` is not set, the server builds WHIP callback URLs from raw \`X-Forwarded-Proto\` and \`X-Forwarded-Host\` headers (\`src/routes/productions.ts:489-499\`). Fastify is initialised without \`trustProxy: true\` (\`src/server.ts:29-36\`), and the code reads raw request headers directly rather than using \`req.protocol\` (which respects \`trustProxy\`). Any client can inject \`X-Forwarded-Proto: http\` and \`X-Forwarded-Host: attacker.com\` to store \`http://attacker.com\` as the WHIP callback URL in CouchDB, redirecting WHIP clients to an attacker-controlled endpoint.

Issue #39 addressed documentation but the raw header read still bypasses Fastify's trustProxy mechanism.

## CVSS Score
6.1 (Medium) — CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:N

## Category
OWASP A04:2021 — Insecure Design

## Affected File(s)
- `src/routes/productions.ts:489-499`
- `src/server.ts:29-36`

```typescript
// Fastify initialised without trustProxy
const fastify = Fastify({ logger: ..., disableRequestLogging: true, bodyLimit: 1_048_576 });

// Raw header read bypasses trustProxy
const proto = (req.headers['x-forwarded-proto'] as string | undefined)?.split(',')[0]?.trim()
  ?? req.protocol ?? 'https';
const host = (req.headers['x-forwarded-host'] as string | undefined)
  ?? (req.headers.host as string | undefined) ?? req.hostname;
return `${proto}://${host}`;
```

## Steps to Reproduce
1. Deploy without `PUBLIC_BASE_URL`
2. `POST /api/v1/productions/<id>/activate` with headers `X-Forwarded-Proto: http` and `X-Forwarded-Host: attacker.com`
3. The stored WHIP callback URL becomes `http://attacker.com/api/v1/...`

## Remediation
**Preferred**: Require `PUBLIC_BASE_URL` in production — throw at startup if `NODE_ENV=production` and `PUBLIC_BASE_URL` is unset.

**Alternative**: Add `trustProxy: true` to Fastify options (trusting only your known reverse-proxy CIDR), then use `req.protocol` and `req.hostname` instead of raw headers. Validate the resulting URL is HTTPS before storing.

## References
- [Fastify trustProxy documentation](https://fastify.dev/docs/latest/Reference/Server/#trustproxy)
- Related closed issue: #39

birme:
Closes #50

## Summary
When `PUBLIC_BASE_URL` is unset, the activation endpoint built the WHIP callback URL (persisted in CouchDB) directly from the raw `X-Forwarded-Proto` / `X-Forwarded-Host` headers, bypassing Fastify's trustProxy mechanism. Any client could inject `X-Forwarded-Host: attacker.com` to persist `http://attacker.com/...` as the WHIP callback endpoint and redirect WHIP clients to an attacker-controlled server (OWASP A04:2021, host-header injection).

Fastify already sets `trustProxy: true` (added for rate-limit IP resolution) and `main.ts` already throws in production when `PUBLIC_BASE_URL` is unset (#39). This PR closes the remaining gap: the fallback path no longer reads raw forwarded headers and no longer persists an unvalidated host.

## Changes
- **`src/routes/productions.ts`**: New exported `resolvePublicBaseUrl(req)` helper + `UntrustedHostError`. Resolution order:
  1. Use `PUBLIC_BASE_URL` if configured (request host is ignored entirely).
  2. Otherwise derive proto/host from trustProxy-aware `req.protocol` / `req.hostname`, then validate:
     - If `TRUSTED_HOSTS` is set, the host must be on the allow-list, else reject.
     - If no allow-list, only loopback/localhost hosts are accepted; any other (proxy-forwarded) host is rejected.
  - Called **before** the production is transitioned to `activating`, so a rejected host returns `400` without leaving the doc mid-activation. Raw `x-forwarded-*` header reads removed.
- **`src/config.ts`**: New optional `TRUSTED_HOSTS` config (comma-separated host allow-list) for request-derived WHIP URLs.

## Test Plan
New `src/__tests__/whip-callback-host.test.ts`:
- [x] Unit: `resolvePublicBaseUrl` prefers `PUBLIC_BASE_URL` and ignores a spoofed host; accepts loopback; accepts allow-listed host; rejects injected non-loopback host with no allow-list; rejects host not on `TRUSTED_HOSTS`.
- [x] Integration: `POST /activate` with `X-Forwarded-Host: attacker.com` returns `400`, and `attacker.com` is never written to CouchDB and the doc is not transitioned to `activating`.
- [x] Integration: legitimate loopback request still returns `200 activating`.

Quality gates (all pass): `pnpm install --frozen-lockfile`, `pnpm run typecheck`, `pnpm run build`, `npx vitest run` (17 files, 164 tests passed — includes 7 new).

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/__tests__/whip-callback-host.test.ts
+++ b/src/__tests__/whip-callback-host.test.ts
@@ -1,0 +1,216 @@
+/**
+ * Security tests for WHIP callback URL host resolution (#50).
+ *
+ * When PUBLIC_BASE_URL is not set, the activation endpoint previously built the
+ * WHIP callback URL directly from the raw X-Forwarded-Proto / X-Forwarded-Host
+ * request headers, bypassing Fastify's trustProxy mechanism. A client could
+ * inject `X-Forwarded-Host: attacker.com` to persist `http://attacker.com` as
+ * the WHIP callback URL in CouchDB, redirecting WHIP clients to an
+ * attacker-controlled endpoint.
+ *
+ * These tests assert that an untrusted / injected host is NOT persisted and NOT
+ * returned — the request is rejected (400) or falls back to the configured URL.
+ *
+ * CouchDB, Strom client, and flow-generator are mocked — no real services required.
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { buildServer } from '../server.js';
+import { resolvePublicBaseUrl, UntrustedHostError } from '../routes/productions.js';
+
+// ---------------------------------------------------------------------------
+// Mock CouchDB
+// ---------------------------------------------------------------------------
+
+const mockGet = vi.fn();
+const mockInsert = vi.fn();
+const mockFind = vi.fn();
+
+vi.mock('../db/index.js', () => ({
+  getDb: () => ({ get: mockGet, insert: mockInsert, find: mockFind }),
+  getSourcesDb: () => ({ get: mockGet }),
+  getOutputsDb: () => ({ get: mockGet }),
+  connectDb: vi.fn().mockResolvedValue(undefined),
+  isDbReady: vi.fn().mockResolvedValue(true),
+  isDbConnected: vi.fn().mockReturnValue(true),
+}));
+
+vi.mock('../ws/controller.js', () => ({
+  default: async () => {},
+  clearAudioState: vi.fn(),
+  clearPipState: vi.fn(),
+  clearFxState: vi.fn(),
+}));
+
+const mockActivateStromFlow = vi.fn();
+const mockDeactivateStromFlow = vi.fn();
+
+vi.mock('../lib/flow-generator.js', () => ({
+  activateStromFlow: (...args: unknown[]) => mockActivateStromFlow(...args),
+  deactivateStromFlow: (...args: unknown[]) => mockDeactivateStromFlow(...args),
+}));
+
+vi.mock('../lib/strom.js', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../lib/strom.js')>();
+  class MockStromClient {
+    system = { version: vi.fn(), iceServers: vi.fn() };
+    flows = {
+      get: vi.fn().mockResolvedValue({ flow: { id: 'flow-abc', running: false } }),
+      start: vi.fn().mockResolvedValue({}),
+      stop: vi.fn().mockResolvedValue({}),
+      delete: vi.fn().mockResolvedValue({}),
+    };
+    mixer = { multiviewEndpoint: vi.fn() };
+  }
+  return { ...actual, StromClient: MockStromClient };
+});
+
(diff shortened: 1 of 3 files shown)


birme (commented):
**APPROVE**

Reviewed as the pr-reviewer agent (independent of the implementer). This closes #50 correctly.

**Vulnerability actually closed.** The fallback path no longer reads raw `X-Forwarded-Proto` / `X-Forwarded-Host` headers. `resolvePublicBaseUrl` now (1) always prefers `PUBLIC_BASE_URL`, ignoring the request host entirely; (2) otherwise derives host from trustProxy-aware `req.hostname` and validates it — rejecting any non-loopback host unless it is on the `TRUSTED_HOSTS` allow-list. `trustProxy: true` is confirmed present (`src/server.ts:69`) and `main.ts:112` already throws in production when `PUBLIC_BASE_URL` is unset, so the layers line up.

**No half-activated doc.** `resolvePublicBaseUrl(req)` is called at `src/routes/productions.ts:638`, BEFORE the `activating` write at line 648. A rejected host throws `UntrustedHostError` (statusCode 400) which is caught and returned as 400, with no CouchDB write and no state transition. The integration test asserts exactly this: `X-Forwarded-Host: attacker.com` -> 400, `attacker.com` never persisted, status never `activating`.

**Validation robustness reviewed, no bypass found.** Host is lowercased before both the loopback and allow-list comparisons (no case bypass); config `trustedHosts` is also lowercased/trimmed. Port is stripped via `host.split(':')[0]` for the checks; the port travels with the same trustProxy-gated forwarded value, so no port-confusion bypass. Trailing-dot / bracketed-IPv6 edge cases fail the allow-list and loopback checks -> reject (fails closed / safe), which is the correct direction.

**Scope & tests.** Only 3 files changed (`config.ts`, `routes/productions.ts`, new `__tests__/whip-callback-host.test.ts`); no `.github/workflows/` changes, no new secrets, no weakened validation elsewhere. New `TRUSTED_HOSTS` env threaded cleanly through `src/config.ts`. I ran the suite locally: 7 new tests pass and the full suite is green (17 files / 164 tests).

Minor (non-blocking, no change required): bracketed-IPv6-with-port hosts and trailing-dot FQDNs are rejected rather than normalized — acceptable since it fails safe and operators can use `PUBLIC_BASE_URL`.

PR #233, opened 2026-09-11
fix(ws): authenticate WebSocket upgrade via header, not query string

### LINKED ISSUE #49: [SECURITY][MEDIUM] API key exposed in WebSocket upgrade query string — visible in proxy logs
## Summary
When \`API_KEY\` is configured, the server accepts authentication via \`?key=\` query parameter on WebSocket upgrade requests (\`src/server.ts:96-98\`). Although Fastify request logging is disabled (\`disableRequestLogging: true\`), reverse proxies (nginx, AWS ALB, Cloudflare), CDNs, and browser DevTools log the full URL including \`?key=<SECRET>\`. Static API keys have no expiry, so log access yields permanent credentials.

## CVSS Score
5.3 (Medium) — CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:N/A:N

## Category
OWASP A02:2021 — Cryptographic Failures

## Affected File(s)
- `src/server.ts:96-98`

```typescript
// WS upgrade carries key in query string since JS WebSocket API doesn't support custom headers
const keyFromQuery = (req.query as Record<string, string>)?.['key'];
const provided = keyFromHeader ?? keyFromQuery;
```

## Steps to Reproduce
1. Set `API_KEY=mysecretkey` on the backend
2. Open browser DevTools → Network tab
3. Connect to `wss://host/api/v1/ws/productions/<id>?key=mysecretkey`
4. The full URL including `?key=mysecretkey` appears in the network log and in nginx/ALB access logs

## Remediation
Replace query-string key passing with a short-lived upgrade token:
1. Add `POST /api/v1/ws/token` (requires Bearer auth) that returns a one-time token valid for 30 seconds
2. The client passes this token as the first WebSocket message
3. The server validates and discards it before routing commands

Alternatively, document that `API_KEY` must be rotated immediately after any log exposure and configure reverse-proxy log redaction for the `key` parameter.

## References
- [OWASP Logging Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Logging_Cheat_Sheet.html)
- [RFC 6455 WebSocket handshake](https://datatracker.ietf.org/doc/html/rfc6455)

birme:
Closes #49

## Summary
When `API_KEY` is configured, the server accepted WebSocket-upgrade auth via a `?key=<secret>` query parameter. Reverse proxies (nginx, ALB, Cloudflare), CDNs, and browser DevTools log the full request URL, so a static, non-expiring key placed there leaks into access logs as permanent credentials (#49, OWASP A02:2021).

This moves API-key auth for the WS upgrade off the query string and onto request **headers**, which are already redacted/omitted from logging.

## Changes
- `src/server.ts`
  - **Removed** the `?key=` query-string auth path entirely (it was the only reference to it in this repo). The key is no longer read from the URL.
  - `Authorization: Bearer <key>` remains supported (REST + non-browser WS clients).
  - **Added** browser-compatible WS auth via `Sec-WebSocket-Protocol`. Browsers can't set arbitrary headers on a WS handshake but can offer subprotocols via `new WebSocket(url, protocols)`. The client offers two subprotocols:
    - `openlive.bearer` — a plain marker
    - `openlive.bearer.<key>` — carries the key
    The server reads the key from the second and echoes back **only** the plain marker (via `handleProtocols`), so the secret is never reflected into the handshake response header either.
  - Added an `extractSubprotocolKey` helper and configured `@fastify/websocket` with `handleProtocols` so the browser handshake isn't torn down for lack of a selected subprotocol.

## Client-compat follow-up
The `?key=` path is **removed**, not deprecated (nothing in this backend repo depended on it). The **`open-live-studio`** browser client connects from a separate repo; if it currently appends `?key=`, it needs a follow-up to switch to:
```js
new WebSocket(url, ['openlive.bearer', `openlive.bearer.${apiKey}`])
```
Filing/handling that studio change is recommended before rollout to any deployment that sets `API_KEY`.

Note: I implemented the header/subprotocol remediation (browser-compatible, no new server state) rather than the issue's alternative short-lived-token suggestion; it directly removes the log-leak vector with a smaller surface and no `/ws/token` endpoint to secure.

## Test Plan
New `src/__tests__/ws-apikey-header-auth.test.ts` drives the WS-upgrade GET through the real onRequest auth hook via `app.inject` (WS controller mocked with a plain handler):
- [x] accepts upgrade with key in `Sec-WebSocket-Protocol` subprotocol
- [x] accepts upgrade with `Authorization: Bearer <key>`
- [x] rejects (401) upgrade carrying the key only in `?key=`
- [x] rejects (401) wrong subprotocol key
- [x] rejects (401) plain marker with no key
- [x] rejects (401) no credentials

Gates (repo root): `pnpm run typecheck`, `pnpm run build` clean; `npx vitest run` → 17 files / 163 tests pass (incl. the 6 new). No new failures.

### DIFF
--- a/src/__tests__/ws-apikey-header-auth.test.ts
+++ b/src/__tests__/ws-apikey-header-auth.test.ts
@@ -1,0 +1,148 @@
+/**
+ * Tests for API-key auth on the WebSocket upgrade (#49).
+ *
+ * Previously the server accepted the API key via a `?key=<secret>` query
+ * parameter on the WS upgrade. Reverse proxies, CDNs, and browser DevTools log
+ * the full request URL, so a static (non-expiring) key placed there leaks into
+ * access logs as permanent credentials.
+ *
+ * The key is now carried in a header instead, and `?key=` is no longer accepted:
+ *   - `Authorization: Bearer <key>` for REST / non-browser WS clients, and
+ *   - the `Sec-WebSocket-Protocol` subprotocol `openlive.bearer.<key>` for
+ *     browser WebSocket clients (which can't set arbitrary headers). The client
+ *     also offers the plain `openlive.bearer` marker, which the server echoes
+ *     back so the secret is never reflected into the handshake response.
+ *
+ * The API-key check lives in the server's onRequest hook (src/server.ts), which
+ * runs for the WS upgrade GET request before the socket is upgraded. We mock the
+ * WS controller with a plain GET handler so `app.inject` can drive the upgrade
+ * request through the real auth hook without needing a live socket: a 401 means
+ * the hook rejected the request, any other status means auth passed and the
+ * request reached the (mocked) handler.
+ *
+ * CouchDB, Strom, and the WS controller are mocked — no real services required.
+ * API_KEY is set before importing the server so config.apiKey (read once at
+ * module load) picks it up.
+ */
+
+import { describe, it, expect, vi, beforeAll } from 'vitest';
+
+const TEST_API_KEY = 'test-secret-key';
+process.env['API_KEY'] = TEST_API_KEY;
+
+const WS_PATH = '/ws/productions/prod-1/controller';
+
+// ---------------------------------------------------------------------------
+// Mock CouchDB
+// ---------------------------------------------------------------------------
+
+vi.mock('../db/index.js', () => ({
+  getDb: () => ({ get: vi.fn().mockResolvedValue(null), insert: vi.fn(), find: vi.fn().mockResolvedValue({ docs: [] }) }),
+  getSourcesDb: () => ({ get: vi.fn(), find: vi.fn().mockResolvedValue({ docs: [] }) }),
+  connectDb: vi.fn().mockResolvedValue(undefined),
+  isDbReady: vi.fn().mockResolvedValue(true),
+  isDbConnected: vi.fn().mockReturnValue(true),
+}));
+
+// ---------------------------------------------------------------------------
+// Mock the WS controller with a plain GET handler on the same path so the auth
+// hook can be exercised via app.inject (no live socket required). Reaching this
+// handler (non-401) proves the onRequest auth hook let the request through.
+// ---------------------------------------------------------------------------
+
+vi.mock('../ws/controller.js', () => ({
+  default: async (fastify: import('fastify').FastifyInstance) => {
+    fastify.get('/ws/productions/:id/controller', async () => ({ upgraded: true }));
+  },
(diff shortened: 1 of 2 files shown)


birme (commented):
**APPROVE**

Verified the change closes #49 cleanly. Reviewed as a separate agent (built + typechecked + ran the new tests locally).

Reasons:
1. **Leak is closed.** The `?key=` query-string path is fully removed from `src/server.ts` — the only remaining `?key=` references are in comments and a test that asserts `?key=` now yields 401. Auth reads solely from `Authorization: Bearer` and the `Sec-WebSocket-Protocol` `openlive.bearer.<key>` subprotocol, both redacted/omitted from logging.
2. **No secret reflection in the handshake response.** `handleProtocols` (`src/server.ts:186-191`) only ever returns the plain marker `openlive.bearer` (or `false`); it never echoes `openlive.bearer.<key>`, so the key cannot leak into the `Sec-WebSocket-Protocol` response header. Confirmed — no defect here.
3. **Auth is sound.** Comparison uses `timingSafeEqual` with a length guard (`src/server.ts:259-262`), so empty/marker-only/wrong keys all reject with 401. The onRequest hook runs before upgrade, so a wrong-key browser client is rejected even though its marker would satisfy `handleProtocols`. Tests cover header-accept, subprotocol-accept, and `?key=`/wrong-key/marker-only/no-cred rejects (6/6 pass; 163 total pass per PR).

Scope is tight (2 files, no `.github/workflows/` changes, no new secrets). The PR body appropriately flags the `open-live-studio` client-compat follow-up (switch to `new WebSocket(url, ['openlive.bearer', 'openlive.bearer.'+key])`) before rollout — the right call versus keeping the insecure default. Ship it once the studio follow-up is filed.

birme:
🤖 daily-backlog-pr Phase 3: independent `pr-reviewer` verdict was **APPROVE** (secure header/`Sec-WebSocket-Protocol` auth, no secret reflection in the handshake, timing-safe comparison, `?key=` removed, tests green). However, merging #232 earlier in this same run advanced `main`, so this PR went `mergeStateStatus: BEHIND` — this repo's branch protection runs required checks in `strict` mode, so an old green run against a prior `main` can't merge. I've refreshed the branch via the Update-branch API to re-queue CI against current `main`. Moving #49 back to **Ready** so the next cycle re-checks this exact open PR and merges it once the re-run check lands (verdict already APPROVE — this is not a re-implementation).

PR #237, opened 2026-09-11
fix(db): always apply safeSelector to db.find() selectors (#64)

### LINKED ISSUE #64: [SECURITY][LOW] safeSelector() Mango injection guard defined but never called — dead-code safety net
## Summary

A `safeSelector` guard function is defined in `src/db/index.ts:10-27` that detects CouchDB Mango query operator keys (`\$in`, `\$or`, `\$regex`, etc.) in user-supplied selectors. However, it is never imported or called anywhere in the codebase. All current `db.find()` calls use hardcoded selector predicates and are not currently vulnerable. The dead code creates false confidence: if a future developer adds a filter endpoint that passes user input to `db.find()` without knowing about this guard, there is no backstop.

## CVSS Score

3.7 (Low) — CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:L/A:N (future risk)

## Category

OWASP A03 — Injection

## Affected File(s)

- `src/db/index.ts:10-27`

## Steps to Reproduce

```bash
grep -rn 'safeSelector' /usercontent/open-live/src/
# Returns only the definition — zero call sites
```

## Remediation

Either:
1. Add a wrapper around `getDb().find()` that always applies `safeSelector` before forwarding to CouchDB, or
2. Add a linter rule / code comment warning that any future `db.find()` with user input must call `safeSelector`, and write a test that exercises it.

Ensuring the function has test coverage prevents it from silently rotting.

## References

- https://docs.couchdb.org/en/stable/api/database/find.html
- https://cheatsheetseries.owasp.org/cheatsheets/Injection_Prevention_Cheat_Sheet.html

birme:
## Summary

Closes #64.

`safeSelector()` was specified as a Mango-injection guard but had zero call sites — a dead-code safety net that would silently rot. This PR defines the guard and wires it onto the `find()` hot path so it becomes a real backstop for any future user-input-driven query, with no behavior change for the existing hardcoded-selector callers.

## Changes

- `src/db/index.ts`:
  - Add `safeSelector()` + `MangoInjectionError`. The guard walks a selector recursively (objects and arrays) and throws on any `$`-prefixed key (`$in`, `$or`, `$regex`, `$gt`, …), which CouchDB treats as a query operator (OWASP A03 — Injection).
  - Intercept `find` in the existing `withTypeGuard` proxy so every `getDb()/getSourcesDb()/getConfigsDb()/getGraphicsDb()/getOutputsDb().find()` call runs the guard on its `selector` before forwarding to CouchDB. All current `.find()` call sites go through these helpers, so the guard is now on the hot path everywhere without touching each caller.
- `src/__tests__/safe-selector.test.ts`: new vitest suite exercising safe pass-through (scalar + nested) and rejection of injected operators (`$in`, `$regex`, `$or`, `$and`), plus null/undefined no-op tolerance — so the guard can't silently rot.

## Test Plan

- [x] `pnpm run typecheck` — clean
- [x] `pnpm run build` — clean
- [x] `npx vitest run` — 248 passed (22 files), including 7 new `safeSelector` tests
- [x] No `.github/workflows/*` changes

## Checklist

- [x] Branched from `main`, not committing to `main`
- [x] No secrets/tokens committed
- [x] Scoped to the issue; no unrelated refactors

### DIFF
--- a/src/__tests__/safe-selector.test.ts
+++ b/src/__tests__/safe-selector.test.ts
@@ -1,0 +1,41 @@
+import { describe, it, expect } from 'vitest';
+import { safeSelector, MangoInjectionError } from '../db/index.js';
+
+describe('safeSelector', () => {
+  it('passes a hardcoded scalar selector through unchanged', () => {
+    const selector = { type: 'production' };
+    expect(safeSelector(selector)).toBe(selector);
+  });
+
+  it('passes a nested selector with no operators through unchanged', () => {
+    const selector = { type: 'source', meta: { tags: ['live', 'studio'] } };
+    expect(() => safeSelector(selector)).not.toThrow();
+  });
+
+  it('rejects a top-level Mango operator key ($in)', () => {
+    expect(() => safeSelector({ type: { $in: ['production', 'source'] } })).toThrow(
+      MangoInjectionError,
+    );
+  });
+
+  it('rejects a $regex operator smuggled into a value', () => {
+    expect(() => safeSelector({ name: { $regex: '.*' } })).toThrow(MangoInjectionError);
+  });
+
+  it('rejects a logical $or operator', () => {
+    expect(() =>
+      safeSelector({ $or: [{ type: 'production' }, { type: 'source' }] }),
+    ).toThrow(MangoInjectionError);
+  });
+
+  it('rejects an operator nested inside an array element', () => {
+    expect(() =>
+      safeSelector({ $and: [{ type: 'production' }] }),
+    ).toThrow(/\$and/);
+  });
+
+  it('tolerates null and undefined selectors (no-op)', () => {
+    expect(() => safeSelector(null)).not.toThrow();
+    expect(() => safeSelector(undefined)).not.toThrow();
+  });
+});
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -5,6 +5,46 @@ import type { ProductionDoc, SourceDoc, ProductionConfigDoc, GraphicDoc, OutputD
 
 let db: Nano.DocumentScope<ProductionDoc>;
 
+/**
+ * Mango-injection guard for CouchDB `find()` selectors. CouchDB treats any
+ * object key beginning with `$` as a query operator (`$in`, `$or`, `$regex`,
+ * `$gt`, …). If user-supplied input is ever merged into a selector without
+ * sanitising, an attacker could smuggle operators to widen a query, trigger an
+ * expensive `$regex`, or bypass an intended `{ type: '...' }` predicate
+ * (OWASP A03 — Injection).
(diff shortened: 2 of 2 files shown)


birme (commented):
## pr-reviewer verdict: APPROVE

Reviewed the source (not just the diff) at head `181e98e`, verified every claim, and ran the checks locally.

### safeSelector implementation (`src/db/index.ts`) — correct
- Recursion is sound: `walk()` handles **arrays** (iterates elements) and **plain objects** (`Object.entries`), and checks `key.startsWith('$')` **before** descending, so a `$`-prefixed key is rejected at any nesting depth — top-level (`$or`, `$and`), nested-in-value (`{name: {$regex: '.*'}}`), and inside array elements are all caught.
- `null`/`undefined`/scalars are no-ops (tolerated), which is the correct behavior for the hardcoded-selector callers.
- No bypass found for the realistic threat model (user input deserialized into a plain object): `Object.entries` enumerates own-enumerable string keys, which is exactly what CouchDB would receive. The guard targets `query.selector` — the correct and only operator-injection surface; `sort`/`fields` are field-name lists, not injection vectors, so not guarding them is correct.

### Hot-path coverage — confirmed on ALL find() call sites
Every DB `.find()` goes through a `withTypeGuard` proxy. I grepped all `.find(` uses: each DB caller binds a **local** `const db = getDb()` / `getSourcesDb()` / `getOutputsDb()` / `getGraphicsDb()` / `getConfigsDb()` (main.ts:29/56, routes/sources.ts, outputs.ts, graphics.ts, production-configs.ts, productions.ts, services/idle-watchdog.ts, services/listener-ports.ts). The raw module-level `db` handle is **not exported** — the only way to reach `.find()` is via the guarded helpers, so no module-local `db.find()` escapes the proxy. (Other `.find(` hits are `Array.prototype.find` on in-memory arrays — unrelated.)

### Existing callers — no behavior change
All current selectors are hardcoded & operator-free (`{type: 'production'}`, etc.), so the guard is a pure no-op for them. `.get()` is unaffected (takes a doc id, not a selector).

### Test quality (`src/__tests__/safe-selector.test.ts`) — meaningful, not trivially green
7 tests cover pass-through (scalar + nested), rejection of `$in`/`$regex`/`$or`/`$and` (top-level, nested, array), and null/undefined no-op. The rejection cases assert `.toThrow(MangoInjectionError)` — they'd fail immediately if the guard were a no-op, so the suite genuinely pins the behavior.

### CI
`build-and-test` — **pass** (18s). Locally: `npx vitest run` on the new suite → 7 passed.

No fabricated claims in the PR body; the "all call sites route through the helpers" claim verified true. Clean, scoped, well-commented backstop. Approving.

PR #240, opened 2026-09-13
fix(audio): number audio channels the way the flow generator wires them

wagenet:
## Summary

The flow generator gives every source an audio mixer channel in mixerInput order, test patterns included, because each one gets a silent `audiotestsrc` (#155). Three other places number the same channels and still skip test patterns:

- `GET /api/v1/productions/:id/audio` in `src/routes/audio.ts`
- `resolveAudioChannelIndex` in `src/ws/controller.ts`, used by `AFV_SET`
- `applyAudioFollow` in `src/ws/controller.ts`, used by CUT, TRANSITION and TAKE

Any test pattern below a real source shifts every real source after it down by one. For sources `__test1__`, SRT, `__test2__`, EFP on `video_in_1..4`, the generator wires SRT to `input_2` and EFP to `input_4`. The route reports SRT as `ch1`, and enabling AFV on EFP routes `ch2_to_main`.

## Fix

I chose one shared function over making `audio.ts` count test sources. With four copies instead of two, patching the three consumers would leave the same drift waiting for the next source type. `src/lib/audio-channels.ts` now holds:

- `assignAudioChannels(assignments, resolve)`: the numbering rule. It sorts by mixerInput, then gives the next channel to each `video_in_N` assignment whose source resolves.
- `loadAudioChannels(assignments)`: resolves virtual sources and sources DB docs, then calls the above. The route and both controller paths use it.
- `VIRTUAL_SOURCES`, moved out of the generator so the route no longer keeps its own copy.

The generator builds its channel map from `assignAudioChannels` and reads it in the per-source loop. The four `audioChannelIndex++` sites are gone, and the generated flow is unchanged.

Two small behaviour changes in the consumers, both toward what the generator already does:
- The `Whip` virtual source resolves before a DB lookup. The controller used to try the DB first.
- Assignments whose mixerInput isn't `video_in_N` get no channel. The generator never wired those.

This matters beyond fader labels. Per-channel sends for guest return feeds (context: #208) would open or close the wrong guest's audio with a wrong index. This fix doesn't depend on that work.

## Tests

New file `src/__tests__/audio-channel-numbering.test.ts`. It generates a real flow for the production above, serves it from a throwaway Strom that the real `StromClient` talks to, and reads each source's expected channel from the generated `input_N` links:

- The generator wires channels 1–4 in mixerInput order. This passed before the fix too and pins the generator side.
- `GET /audio` returns the right `mixerInput` and label for each `chN`.
[...]

PR #220, opened 2026-09-09
docs(spec): integrated guest calling (WHIP/WHEP + Open Intercom)

birme:
Proposed architect spec for epic #208 — browser guest join via WHIP/WHEP with Open Intercom talkback. Docs-only (`docs/specs/guest-calling-intercom.md`); no product code.

## Summary
- Baselines a **WHIP-video + intercom-audio-talkback** topology so the epic does NOT hard-depend on unmerged SVT intercom-video work; talkback degrades cleanly when intercom is not configured.
- Guest = WHIP source assigned to a mixer input, reusing the existing `/whip/:mixerInput` and `/whep-proxy` routes; take-to-air reuses existing vision-mixer switching.
- New invite/join/kick REST endpoints, `GuestInviteDoc`/`GuestSessionDoc`, and a `GUEST_STATE` WS event.
- Notes that accurate `on-air` guest state depends on #209's contribution tally.

## Open Questions (genuine product/topology decisions)
1. Guest-video topology: intercom-video (needs SVT merge — confirm location/timeline) vs WHIP-video + audio talkback (proposed baseline).
2. Per-guest mix-minus at launch vs program-with-delay for v1.
3. Invite auth model (proposed: signed expiring hashed tokens).
4. Automated per-production intercom line provisioning vs manual v1.
5. Max simultaneous guests / Strom flow sizing.
6. Green-room preview surface (Studio ticket).

Likely needs breakdown into sub-issues after the topology decision. Marked **Proposed**. No `Closes #208`.

Generated with Claude Code (architect agent).

Issue #208, opened 2026-09-05
Integrated guest calling: browser guest join with WHIP/WHEP and Open Intercom talkback


svensson00:
## Summary

Add an integrated remote-guest ("guest calling") workflow to Open Live: a guest joins from a plain browser link, contributes camera/microphone over WHIP, receives a return feed over WHEP, and can be previewed and spoken to by the operator — via [Open Intercom](https://github.com/Eyevinn/intercom-manager) — before being taken on air.

## Motivation

Open Live supports WHIP sources and REMI-style remote production, but there is no managed guest workflow: no join link, no green room, no return feed to the guest, and no operator↔guest talkback. Newsroom-style and emergency-broadcast workflows typically require that an operator can preview and communicate with a remote contributor before placing the contributor on air. Eyevinn already ships the missing half as a product — Open Intercom (intercom-manager + intercom-frontend), which itself exposes WHIP/WHEP — so this epic is primarily an integration between the two product families.

## Proposed scope

- **Guest join flow**: production-scoped invite link; guest joins from a standard browser on desktop or mobile, no specialist software; guest's camera/mic published over WHIP into the production as a source.
- **Green room / preview**: the guest source is previewable (multiviewer / preview bus) before being placed on air.
- **Return feed**: guest receives program (or a mix-minus of it) over WHEP, so remote participation is usable for live interviews.
- **Talkback via Open Intercom**: each Open Live production gets an associated intercom line (or lines); operator can talk to a guest pre-air and during production. Reuse intercom-manager's session/line model and intercom-frontend components rather than building new comms.
- **Video in the intercom path — to confirm during spec**: video support for Open Intercom has been implemented by SVT but may not yet be merged to the main track. The spec must (1) confirm where that work lives and its upstreaming timeline, and (2) decide the guest-video topology: video via the intercom line (if/when merged) vs. video via Open Live's existing WHIP source path with intercom carrying talkback audio only. The epic should not hard-depend on unmerged work: the WHIP-video + audio-talkback fallback must be a first-class outcome.
- **Control surfaces**: REST + WebSocket parity with the rest of Open Live (create invite, list/kick guests, guest state events: joined, previewing, on-air, left), so external automation can drive guest management.
- **Studio UI**: guest list with preview, talk button, and take-to-air — exact surface to be defined in the spec.

## Out of scope

- PSTN/phone dial-in and conferencing gateways (Teams/Zoom) — possible follow-ups.
- Recording of guest calls.
[...]

PR #241, opened 2026-09-13
docs(spec): design the guest return feed (guest-calling OQ2)

wagenet:
Answers Open Question 2 of the guest-calling spec (epic #208): does the WHEP return feed need per-guest mix-minus at launch? Docs only, in `docs/specs/guest-calling-intercom.md`. The spec stays **Proposed**.

## What the spec now says

- **Mix-minus at launch**, default `program-minus`, with `program` as a live per-guest switch.
- **One post-fader aux bus per guest** on `builtin.mixer`, and one WHEP output per guest carrying program video plus that single return track.
- **Return routes** under `/productions/:id/returns/:mixerInput/...`, `RETURN_SET`/`RETURN_STATE`, and an optional `ProductionSourceAssignment.returnFeed`.
- **`low-latency-minus`** (for guests talking to each other on air) is described, but not in v1.

## Changes to what #220 merged

- **`whepReturnUrl` is gone** from the join response, replaced by `feeds`, `modes`, `defaultMode` and `returnMode`. It pointed at `/whep-proxy?target=`, which forwards to any URL on the Strom host, so it could not be limited to one guest's token.
- **Open Question 2** now asks what is still open: which mix feeds a low-latency return. **Open Question 7** is new: the guest picture source.
- **Two new risks.** Audio channel numbering disagrees between the flow generator and the controller when test sources are assigned, so a return could close the wrong send; #240 fixes that separately. And WHIP PATCH/DELETE accept any session URL on the Strom host.

## Decisions for a maintainer

- **Guest picture source (OQ7).** The spec proposes one shared low-bitrate encode for guests, over passing the program encode through or encoding per guest.
- **Mix behind a low-latency return (OQ2).** An aux bus keeps each guest's gate and compressor but shares the audio mixer's latency. `builtin.liveaudiorouter` avoids that latency but carries raw microphones and has no limiter unless Eyevinn/strom#795 lands. The spec sets out both and doesn't pick one.

<details><summary>Rationale</summary>

**Why mix-minus, not program-with-delay.** A guest who hears their own voice come back a few hundred ms late struggles to keep speaking, so a plain program return fails the live interview the epic is for.

**Why on-air conversation is not intercom.** The return feed is what the audience hears; intercom is off-air coordination. Holding an on-air conversation over intercom would make each guest hear the others twice (live on intercom, then late in the return), and none of the producer's program moves, such as ducking, would reach the intercom line.

**Why an aux bus, not `liveaudiorouter`, for the synced modes.** The router has no access to post-fader channel audio, so it cannot follow faders, mutes or channel processing. Aux send levels are live, so the mode switches mid-show with one property change.

**Why one switched track per guest.** The shared WHEP outputs are capped at 8 audio tracks and send every track to every viewer, so per-guest returns cannot ride them. One switched track, rather than a track per mode, avoids depending on whether browsers lip-sync a second audio track to video, which is untested.

**Why a shared low-bitrate picture encode.** A guest's picture is a confidence monitor. With pre-encoded input, Strom's `whep_output` passes the stream through, so webrtcsink has no encoder for congestion control to act on (inferred from its design, not tested), and a remote guest receives program bitrate. Join time also follows program's GOP: webrtcsink's codec discovery needs a keyframe and Strom recommends a one-second GOP for WebRTC, while `builtin.videoenc` defaults to 60 frames. A dedicated guest encode gets a predictable bitrate and a GOP chosen for fast joins without touching program. Per-guest encoding adapts, but not on macOS VideoToolbox, where webrtcsink logs "Bitrate handling is not supported yet".

**Why the fast feed is a separate session.** On a test rig, a receiver given audio and video in one session delayed the audio to lip-sync it, which restores the delay the mode exists to remove.

</details>

<details><summary>Evidence for the low-latency questions</summary>

A local test rig (five mix-minus seats, 127 trials, tone in to tone out through Strom, loopback) found:

- End-to-end delay tracked the WHIP ingest jitterbuffer almost one-for-one from 20 to 400 ms. Strom's default is 400 ms, set at build time, and Open Live never sets it.
- Receiver-side webrtcbin `latency` from 20 to 1000 ms made no difference.
- A negative `ts_offset_ms` on the return outputs made no difference, but that rig's paths differed by a few milliseconds, so there was no gap to recover. Open Live's flow, where program audio waits on `mix_latency` and the slowest SRT source, has not been tested. The check is one trial: a conservative negative offset on a router-fed return, in a production with a slow SRT source.
- `liveaudiorouter` added about 53 ms at its default 30/30/10 and about 20 ms at 15/10/5.
- A four-voice N-1 return peaked at 0 dBFS with every crosspoint at unity, so either mix needs a limiter. Aux buses have none, and the router's would come from Eyevinn/strom#795.

If an aux bus proves too slow and the router too raw, the fix is in Strom: the mixer exposes no processed per-channel pad, and an aux bus used as a direct out still carries the mixer's aggregation latency.

</details>

Strom behaviour was checked against Eyevinn/strom `main` at `eae2467`, open-live references against `137ef3d`.


birme (approved):
Phase 1.5 spec-review gate (team-lead) — APPROVED.

Reviewed `docs/specs/guest-calling-intercom.md` against epic-workflow.md Phase 1 checklist: API contract (guest invites/join, per-guest return feeds, WS GUEST_STATE/RETURN_STATE events, error-code table) ✓, data model + migration (GuestInviteDoc/GuestSessionDoc, additive ProductionSourceAssignment.returnFeed, no CouchDB migration) ✓, Mermaid sequence diagram ✓, configuration/env vars ✓, open questions present ✓.

The three blocking open questions were decided by the PM decision authority @svensson00 on the #208 epic thread (2026-09-16, https://github.com/Eyevinn/open-live/issues/208#issuecomment-5696583364) and are now recorded inline in the spec:

- OQ1 (guest-video topology): guest video goes through Open Live's own WHIP-in path; Open Intercom carries talkback audio only. Video-over-intercom is out of scope for this epic, so the unmerged-SVT-intercom-video dependency risk is closed.
- OQ2 (return feed): per-guest mix-minus (`program-minus` default) — exactly what this spec's Return feed design specifies (one post-fader aux bus per guest).
- OQ7 (guest picture source): v1 guests see program output via WHEP; per-guest return video is a later enhancement.

Status flipped Proposed → Accepted. The remaining open questions (low-latency-return mix, invite auth model, intercom line provisioning, capacity limits, green-room preview location) are implementation-detail or dependent-ticket calls and do not block accepting the spec or breaking the epic into sub-issues. team-lead reviewing architect's spec is not self-review.

PR #242, opened 2026-09-14
fix(flow-generator): no 125 ms mixer floor without SRT/EFP sources

wagenet:
Written by Claude (Anthropic's coding agent) on behalf of @wagenet.

With no SRT/EFP source assigned, the generator set `min_upstream_latency: 125` on both mixers. That value is a floor on the latency a mixer assumes its inputs report. WHIP slots report almost none (the WHIP jitterbuffer runs in the session pipeline, before Strom restamps buffers into the flow), and test and HTML sources report little. So a WHIP-only production paid the full floor in program delay. This change falls back to `builtin.mixer`'s default of 30 ms. Productions with SRT/EFP sources still get the highest source latency.

## Measured

Strom `upstream/main` eae2467 (release build). The flow was the one `activateStromFlow` generates for two WHIP sources; only the vision mixer was forced to `compositor_preference: cpu`, because the GL path cannot link to `videoenc` on macOS. Each trial used a fresh backend. Publishers sent Opus tick tones plus H.264 video over loopback. Delay is tick-in to tick-out on the `pgm` WHEP audio, captured in one GStreamer pipeline. Three trials per arm, arms alternated, host load average 19–35.

| mixers `latency` / `min_upstream_latency` | Strom-reported pipeline latency | program audio delay (median) |
|---|---|---|
| 100 / 125 (before) | 475 ms | 932–954 ms |
| 100 / 30 (this PR) | 380 ms | 847–855 ms |

- **Continuity:** no PTS gaps or dropouts ≥20 ms in either arm.
- **Moderate network jitter** (0–20 ms per packet): no dropouts at mixers 30/30.
- **SRT source present:** the floor made no difference. With an SRT source (latency 125 or 400), `min_upstream_latency` = max SRT latency and = 30 gave identical reported latency (475 / 750 ms) and delay. `srtsrc` reports its own latency, so the SRT branch is kept as-is.

Mixer `latency` (`mix_latency`, default 100) is left alone: it is a Studio production setting. It costs about 3× its value in program delay, because the monitor bus chains three aggregators (aux → solo → monitor) and `monitor_out` feeds the WHEP outputs.

## Tests

- `pnpm exec vitest run`: 250 passed.
- `pnpm typecheck`: clean.
- The new "no SRT/EFP source" test fails on `main` (`expected [125, 125] to deeply equal [30, 30]`); the SRT/EFP test passes on both.

Not tested: real remote WHIP publishers, and HTML (`cefsrc`) sources.


birme (commented):
## PR Review: fix(flow-generator): no 125 ms mixer floor without SRT/EFP sources

### Verdict: APPROVE

Scoped, correct, well-tested bug fix. Every field name, value, and branch condition claimed in the description is verified against the actual source on the PR branch. No fabrication, no scope creep, no security concern.

### Critical Issues (must fix before merge)
- None.

### Warnings (should fix)
- None.

### Nits (optional)
- The per-source SRT fallback `s.latency ?? 125` (flow-generator.ts:429) still uses the old `125` literal, while the no-source case now uses the named `STROM_MIXER_MIN_UPSTREAM_LATENCY_MS` (30). This is intentional and correct — an explicitly-configured SRT/EFP source with no `latency` set should keep the 125 ms assumption, distinct from the "no real sources at all" default. Left as a nit only in case a future reader wonders why the two numbers differ.

### What Was Verified
- **`min_upstream_latency` is real:** set on both mixers at `src/lib/flow-generator.ts:435` (`builtin.vision_mixer`) and `:441` (`builtin.mixer`). Not fabricated.
- **`builtin.mixer` / `builtin.vision_mixer` block IDs are real:** mixer blocks located at flow-generator.ts:248-256; same IDs the new test's `mixerFloors()` helper queries. Not fabricated.
- **The 125 → 30 change is precise:** the only behavioral change is the no-SRT/EFP fallback: `maxSourceLatency = srtLatencies.length > 0 ? Math.max(...srtLatencies) : STROM_MIXER_MIN_UPSTREAM_LATENCY_MS` (flow-generator.ts:430), the new constant = 30 (flow-generator.ts:32). Not fabricated.
- **SRT/EFP branch genuinely unchanged:** the `srtLatencies` filter (flow-generator.ts:428) excludes `test1`/`test2`/`whip`/`html`, so any real SRT/EFP source still drives `Math.max(...)`. When sources exist the `? :` takes the max branch exactly as before. Confirmed.
- **`mixLatency` / `latency` left alone:** `props['latency'] = mixLatency` (flow-generator.ts:434, :440) is untouched by the diff — matches the description's claim that the Studio `mix_latency` setting is not changed.
- **Test coverage hits the changed path:** virtual sources map `'Whip' → streamType 'whip'` and `'__test1__' → 'test1'` (flow-generator.ts:64-65), both excluded by the filter, so the "no SRT/EFP source" test yields an empty `srtLatencies` and asserts `[30, 30]` — exercising the new fallback. The second test uses real `srt`/`efp` docs (latency 200) and asserts `[200, 200]`, guarding the unchanged branch. Both are correct.
- **Tests pass on branch:** `vitest run src/__tests__/flow-generator.test.ts` → 5 passed (the 2 new + 3 existing).
- **Scope:** diff touches only `src/lib/flow-generator.ts` and `src/__tests__/flow-generator.test.ts` (+64/-2). No unrelated logic, no config, no auth/route changes.
- **Security:** external contributor (@wagenet). No secrets, no hardcoded credentials, no weakened validation, no injection vectors, no backdoor. The CouchDB mock change is test-only (adds a source-doc lookup map). Clean.

Note: CI is green but the branch is BEHIND main (stale run); branch refresh is being handled separately by the orchestrator. This verdict is the code read-through only.

birme:
🤖 Phase 3.5 orphan-PR housekeeping: read-through verdict is **APPROVE** (full review posted above — flow-generator field names, values, and the SRT/EFP branch condition all verified against source; no scope creep; no security concern from external contribution).

Not merged this run because the branch was `BEHIND` main and this repo runs `required_status_checks` in `strict` mode, so the existing green CI ran against stale `main`. I've triggered an **Update branch** (server-side merge of current `main`), which re-queues CI. Once that check lands green, the next automation run will merge this via the existing-open-PR re-entry path — no re-review needed unless the diff changes.

birme (commented):
APPROVE

Independent review (external contributor, extra scrutiny). Change is minimal, correct, and well-scoped.

**What it does (flow-generator.ts:463-475):** The only behavioural change is the empty-list fallback for `maxSourceLatency`, moved from a hardcoded `125` to a named constant `STROM_MIXER_MIN_UPSTREAM_LATENCY_MS = 30` (builtin.mixer's own default). Applied identically to both mixers via the existing `min_upstream_latency` assignments (lines 480, 486).

**Verification of the stated claims:**
- (a) SRT/EFP path unchanged: the `srtLatencies` filter still excludes `test1/test2/whip/html`, keeps SRT/EFP, and maps to per-source `s.latency ?? 125`. Per-source default and `Math.max` semantics are untouched. ✅
- (b) Fallback only fires when `srtLatencies.length === 0` (no SRT/EFP source). ✅
- (c) New tests assert `[30,30]` for WHIP+test and `[200,200]` for an SRT(200)+EFP+WHIP mix (EFP defaults to 125, so max=200 — asserts SRT/EFP behaviour is preserved, not just [125,125]). ✅
- (d) No scope creep: `git diff --stat` = 2 files, +64/-2. `mix_latency`/`latency` deliberately left alone. ✅
- (e) Nothing suspicious: no backdoors, no weakened validation, no hardcoded secrets, no network/auth touch. Test mock upgrade (`vi.hoisted` source-doc map) is standard and isolated. ✅

**Local results (fresh clone of pr242):** `pnpm typecheck` clean · `pnpm build` clean · `vitest run` 515/515 passed (flow-generator: 5/5, incl. both new cases).

Well-reasoned, measured, and behaviour-preserving for SRT/EFP. No defects found. Missing `Closes #N` is the only nit — cosmetic, not blocking.

birme (approved):
Independent pr-reviewer verdict: APPROVE. Minimal, correct, well-scoped WHIP-latency-floor fix; SRT/EFP path unchanged; local typecheck/build/vitest (515/515) all green. Merging.

Issue #244, opened 2026-09-14
docs: README and controller-websocket.md have drifted from src on five points, two security-relevant


alexbj75:
## Problem

Five documented facts in `README.md` and `docs/controller-websocket.md` no longer match `src/`. Each was verified against the source, not inferred.

| # | Docs say | Source says | Where |
|---|---|---|---|
| 1 | WebSocket clients pass the key as `?key=` | **`?key=` is no longer accepted.** Clients offer `Sec-WebSocket-Protocol: openlive.bearer` plus `openlive.bearer.<key>`; the server extracts via `extractSubprotocolKey()` and echoes back only the plain marker | `src/server.ts` (the comment cites #49), `src/__tests__/ws-apikey-header-auth.test.ts` |
| 2 | `CORS_ORIGIN` defaults to `http://localhost:5173` | `process.env['CORS_ORIGIN'] ?? undefined`, then `origin: false` with a `[security]` warning. Accepts a comma-separated list or `*` | `src/config.ts`, `src/server.ts`, `src/__tests__/cors-default-origin.test.ts` (guards #54) |
| 3 | `COUCHDB_NAME` is a configurable variable with a default | **Inert.** The name is hardcoded and the variable appears nowhere in `src/` | `src/db/index.ts` |
| 4 | `STROM_TOKEN` is the token variable | Legacy fallback. `STROM_AUTH_TOKEN ?? STROM_TOKEN`, and `STROM_AUTH_MODE` (`osc` default, `direct` for self-hosted Strom) is undocumented in the README | `src/config.ts` |
| 5 | Environment variable table omits `PUBLIC_BASE_URL`, `SRT_PASSPHRASE_KEY`, `TRUSTED_HOSTS`, `COUCHDB_USER`/`COUCHDB_PASSWORD`, `STROM_PORT_LEASE_CLIENT_ID` | `PUBLIC_BASE_URL` is a **hard production startup guard** and a host-header-injection control; `SRT_PASSPHRASE_KEY` fails closed on the encrypt/decrypt path | `src/main.ts`, `src/lib/srt-passphrase-crypto.ts`, `src/config.ts` |

[...]

PR #248, opened 2026-09-14
docs: correct Strom env vars and drop OSC-hosted Strom as supported

LucasMaupin:
## Env var corrections

The README's env table was wrong in two ways:

- It documented `STROM_TOKEN`. That still works as a legacy alias, but `.env.example` and `config.ts` use **`STROM_AUTH_TOKEN`**.
- `STROM_AUTH_MODE` was missing entirely, despite defaulting to `osc` — so anyone following the README against a self-hosted Strom would get SAT exchange attempted on a plain Bearer token.

Added a table covering `direct` vs `osc`.

## OSC-hosted Strom is no longer presented as supported

`eyevinn-strom` removed from "Required services", since Strom is not deployed as an OSC service instance — `STROM_URL` points at a GPU host outside the catalogue.

**The `osc` auth-mode code is unchanged and stays.** It is ~60 lines, it works, and it is the only way to reach a Strom behind OSC ingress — useful as an integration-test target. This is a docs-only change to stop presenting the catalogue service as a supported production backend.

Context: a full stack was stood up against a catalogue `eyevinn-strom` instance to check. The auth path worked first try (`{"db":true,"strom":true}`), but the instance has no GPU (software HEVC for PGM + multiview, queues at 13s within a minute), WHEP returned 502 from an internal proxy failure, and with `DatabaseUrl` unset flows are stored on ephemeral disk and lost on every restart.


birme:
Thanks for this @LucasMaupin. Flagging two things from the daily-backlog-pr housekeeping sweep:

1. **Overlap with just-merged #247.** PR #247 (Closes #244) merged into `main` earlier today and already renames `STROM_TOKEN` → `STROM_AUTH_TOKEN` (legacy alias noted) and documents `STROM_AUTH_MODE` in the env table. That's why this branch now shows as conflicting. Please rebase on current `main` — the env-table portion of this PR is now largely redundant, though your **`STROM_AUTH_MODE` behaviour table** (`direct` vs `osc`) is a nice addition #247 didn't include and is worth keeping.

2. **Support-policy claim needs a maintainer decision.** The change stating that OSC-hosted Strom (`eyevinn-strom` catalogue service) is **not a supported production backend**, and removing it from the required-services list, is a product/support-policy statement rather than a source-drift correction — this automation can't verify or merge it unilaterally. @svensson00 can you confirm whether OSC-hosted Strom should be documented as unsupported (integration-testing only)? Once that's decided and the branch is rebased, the docs portion can go in.

Leaving this open and flagging for human triage; not merging or closing it automatically.

LucasMaupin:
Closing — the env var corrections here were superseded by #247, which landed the same STROM_AUTH_TOKEN / STROM_AUTH_MODE fixes. Will re-open a narrower PR if the OSC-support change is still wanted separately.

svensson00:
For what its worth, I can confirm that OSC-hosted Strom should be documented as unsupported (integration-testing only) for now

Issue #78, opened 2026-06-29
[SECURITY][MEDIUM] srtUrl() validator has no private IP blocklist — SRT SSRF to internal hosts via GStreamer pipeline


LucasMaupin:
## Summary

The `srtUrl()` validation function in `src/lib/url-validation.ts` checks that a URI starts with `srt://` and contains no shell-dangerous characters, but does **not** check whether the hostname resolves to a private or loopback address. An authenticated user (or any user when `API_KEY` is unset — see #56) can register a source or output pointing at `srt://127.0.0.1:5005`, `srt://10.0.0.1:5005`, or `srt://192.168.x.x:5005`. These are passed as `srt_uri` properties directly to GStreamer via Strom, enabling SSRF to services on the same host or internal network from within the GStreamer media pipeline. The earlier fixes in #18 and #23 added scheme and injection validation but did not add an IP range blocklist.

## CVSS Score
6.5 (Medium) — CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:N/A:N

## Category
OWASP A10:2021 — Server-Side Request Forgery (SSRF)

## Affected File(s)
- `src/lib/url-validation.ts` — `srtUrl()` function (scheme + dangerous-char check only)
- `src/routes/sources.ts` — SRT source address passes through `srtUrl()`
- `src/routes/outputs.ts` — SRT output address passes through `srtUrl()`

## Steps to Reproduce
1. `POST /api/v1/sources` body: `{"name":"ssrf-test","address":"srt://127.0.0.1:9999?mode=caller","streamType":"srt"}`
2. Observe: passes `srtUrl()` validation.
3. Activate a production with this source — GStreamer (via Strom) attempts a caller connection to `127.0.0.1:9999` on the host running Strom.

The same applies to private LAN ranges (`10.x.x.x`, `172.16.x.x`, `192.168.x.x`) and IPv6 loopback (`::1`).

## Remediation

Extend `srtUrl()` to reject private and loopback address ranges after parsing the hostname:

```typescript
import { isPrivateIP } from './ip-utils'; // shared helper, see below

function srtUrl(value: string): string {
  if (!value.startsWith('srt://')) throw new Error('Must be srt:// URL');
  const dangerous = /[!;\s]/.test(value);
  if (dangerous) throw new Error('URL contains unsafe characters');

  // Allow listener mode (empty host): srt://:6000?mode=listener
  const parsed = new URL(value.replace('srt://', 'http://'));
[...]

PR #251, opened 2026-09-15
docs(spec): configurable SRT private-host allowlist (#246)

### LINKED ISSUE #246: Make SRT private/loopback/link-local rejection deployment-configurable (self-hosted on-prem encoders)
## Context

Split out from #245. `srtUrl()` (`src/lib/url-validation.ts`) unconditionally rejects private, loopback, and link-local SRT targets via `isPrivateHost()`. This is the intended SSRF hardening from #78 and is **correct for cloud deployments** — but it is a blanket, deployment-agnostic block with no allowlist or opt-in.

## Problem

Self-hosted / on-prem live-production topologies routinely point an SRT source at a private-range encoder on the same LAN (e.g. `srt://10.x.x.x:9999`). Those legitimate setups are now permanently blocked at source creation with:

> `SRT URL must not target private, loopback, or link-local addresses`

## Proposed enhancement

Provide a deployment-configurable escape hatch that keeps the cloud default safe:

- A boolean like `ALLOW_PRIVATE_SRT` (default `false`), **or**
- A `TRUSTED_SRT_HOSTS` / CIDR allowlist so operators can permit specific private ranges.

The SSRF fix from #78 must remain the default; this only relaxes it for operators who explicitly opt in on a trusted network.

## Notes

- Architect flagged this needs its own spec (config surface + how it interacts with the existing `isPrivateHost()` guard). Filing to the board Backlog for triage.
- Do **not** revert #78.

Follow-up from #245.

birme:
## Summary

Spec for #246 — a design spec (docs-only, Phase 1 architect output) for making the SRT private/loopback/link-local rejection deployment-configurable for self-hosted / on-prem encoder setups.

- Recommends a host/CIDR allowlist `TRUSTED_SRT_HOSTS` (default empty) over a coarse `ALLOW_PRIVATE_SRT` boolean, so the cloud default stays safe and only explicitly-listed ranges are reopened.
- Mirrors the existing `TRUSTED_HOSTS` split/trim/filter pattern in `src/config.ts`.
- The #78 SSRF hardening remains the default and is NOT reverted; the allowlist can only relax a private-range rejection, never bypass format/control-char/length guards or affect http/graphic SSRF checks.
- Grounded in real code: `srtUrl()` / `isPrivateHost()` in `src/lib/url-validation.ts`, callers in `src/routes/sources.ts` and `src/routes/outputs.ts`, config patterns in `src/config.ts`.

Spec file: `docs/specs/configurable-srt-private-host-allowlist.md`

## Scope

Docs-only — touches a single file under `docs/specs/`. No implementation.

## Test plan

- [ ] Team-lead spec review (Phase 1.5)
- [ ] Confirm design decision (allowlist over boolean) before implementation

Note: this spec does not close #246 — implementation is a separate PR.


birme:
Phase 1 spec-review GATE (team-lead) — **APPROVED**.

(Recorded as a comment: the formal `--approve` review is blocked because this PR was authored under the same GitHub identity the gate runs as — GitHub disallows self-approval. This is a tooling/identity constraint, not a spec gap.)

Docs-only confirmed: single added file `docs/specs/configurable-srt-private-host-allowlist.md` (256 additions, no implementation touched).

Checklist outcomes:
- **API / validation contract** — precise. Exact `srtUrl(url, trustedSrtHosts=[])` signature, the modified final gate (`hostname && isPrivateHost(hostname) && !isTrustedSrtHost(...)`), the new error hint text, and the precedence (format/control-char/length guards stay mandatory and non-overridable; the allowlist only relaxes a private-range rejection, never bypasses format checks or affects public hosts). Composition with `isPrivateHost()` and preservation of #78 as the default are correct.
- **Data model / config** — reuses the existing `TRUSTED_HOSTS` idiom verbatim (`src/config.ts:82-85`): new `trustedSrtHosts` / `TRUSTED_SRT_HOSTS`, `string[]`, default `[]` (= today, #78 preserved), same split/trim/lowercase/filter parsing. No CouchDB schema change.
- **Sequence diagram** — present (mermaid validation-path flowchart).
- **Configuration section** — present, with cloud-default / single-encoder / subnet / multi-range examples and an explicit security note against metadata/wildcard ranges.
- **Open Questions** — all resolved in-spec (allowlist vs boolean, CIDR vs bare IP, config injection, listener form). Only a non-blocking optional follow-up (startup warning log) left to implementer discretion. No "TBD, needs a human".

Code-ref sanity check against `main`: verified `url-validation.ts:237-239` (quoted gate), `isPrivateHost()` `:34-60`, `srtUrl()` pure config-free at `:211`, `config.ts:82-85` idiom, callers `sources.ts:30,61,167` / `outputs.ts:20,107`. Every cited line matches — no invented fields, no misdescribed guard.

Sound spec; proceeding to merge.

PR #253, opened 2026-09-15
fix(macros): validate macro actions with a discriminated union

### LINKED ISSUE #53: [SECURITY][LOW] Macro action Zod schema uses flat optional fields instead of discriminated union
## Summary
\`MacroActionSchema\` (\`src/routes/macros.ts:7-15\`) defines all action fields as a flat object with optional fields, permitting cross-type field mixing — e.g. a \`CUT\` action can include \`overlayId\` and \`durationMs\` that belong to \`GRAPHIC_ON\` and \`TRANSITION\`. The WebSocket controller (\`src/ws/controller.ts:72+\`) correctly uses \`z.discriminatedUnion\`, but macros stored in CouchDB and replayed via \`MACRO_EXEC\` pass through the looser schema. If the controller's replay logic changes, extraneous fields could trigger unintended Strom calls.

This is a defence-in-depth issue; no direct exploit exists in the current code.

## CVSS Score
3.1 (Low) — CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:N/I:L/A:N

## Category
OWASP A03:2021 — Injection (schema confusion)

## Affected File(s)
- `src/routes/macros.ts:7-15`

```typescript
// Current: flat, any combination of optional fields allowed
const MacroActionSchema = z.object({
  type: z.enum(['CUT', 'TRANSITION', 'TAKE', 'GRAPHIC_ON', 'GRAPHIC_OFF', 'DSK_TOGGLE']),
  sourceId: z.string().optional(),
  transitionType: z.string().optional(),
  durationMs: z.number().int().positive().optional(),
  overlayId: z.string().optional(),
  layer: z.number().int().min(0).optional(),
  visible: z.boolean().optional(),
});
```

## Remediation
Replace with a discriminated union matching the pattern in `controller.ts`:
```typescript
const MacroActionSchema = z.discriminatedUnion('type', [
  z.object({ type: z.literal('CUT'), sourceId: z.string() }),
  z.object({ type: z.literal('TRANSITION'), sourceId: z.string(), transitionType: z.string(), durationMs: z.number().int().positive().optional() }),
  z.object({ type: z.literal('TAKE') }),
  z.object({ type: z.literal('GRAPHIC_ON'), overlayId: z.string() }),
  z.object({ type: z.literal('GRAPHIC_OFF'), overlayId: z.string() }),
  z.object({ type: z.literal('DSK_TOGGLE'), layer: z.number().int().min(0), visible: z.boolean().optional() }),
]);
```
This aligns the stored macro schema with the validated WS schema and prevents schema drift.

birme:
Closes #53

## Summary

`MacroActionSchema` in `src/routes/macros.ts` defined all action fields as one flat object with every field optional, so a `CUT` action could illegally carry `overlayId`/`durationMs` (which belong to `GRAPHIC_ON`/`TRANSITION`). Macros stored in CouchDB and replayed via `MACRO_EXEC` passed through this looser schema, while the WS controller (`src/ws/controller.ts`) already validates its inbound messages with `z.discriminatedUnion`. This is a defense-in-depth / schema-confusion gap (OWASP A03) with no live exploit.

This PR replaces the flat schema with a strict discriminated union on the action `type`, mirroring the controller's pattern, so cross-type field mixing is now a hard 400.

## Changes

- `src/routes/macros.ts`: Replace flat optional `MacroActionSchema` with a `z.discriminatedUnion('type', [...])`. Each variant declares only the fields its `MACRO_EXEC` replay path actually consumes (`CUT`: `sourceId`; `TRANSITION`: `sourceId`/`transitionType`/`durationMs`; `TAKE`: none; `GRAPHIC_ON`/`GRAPHIC_OFF`: `overlayId`; `DSK_TOGGLE`: `layer`/`visible`). Per-field constraints are preserved exactly from the old schema (no validation loosened), and field optionality is unchanged so previously-valid single-type payloads still pass. Each variant is `.strict()` because Zod otherwise silently strips unknown keys instead of rejecting them — strict mode is what makes cross-type mixing an actual 400.
- `src/__tests__/macros.test.ts`: Updated one existing fixture that mixed `GRAPHIC_ON` with DSK fields (`layer`/`visible`) to a clean `GRAPHIC_ON`. Added a focused describe block asserting each valid variant still returns 201 and that cross-type combinations (e.g. `CUT` + `overlayId`, `CUT` + `durationMs`, `GRAPHIC_ON` + DSK fields, `TAKE` + `sourceId`, unknown type) are rejected with 400 and never reach the DB.

## Design note

The union follows the macro executor's real field usage (`sourceId`, plain-string `transitionType`), which differs from the controller's `mixerInput`/`TransitionTypeSchema` shape — the two schemas validate different wire formats. The discriminated-union *technique* is mirrored, not the exact field names, so accepted valid macro payloads are unchanged.

## Test Plan

Ran all code-quality checks from repo root; all green:

- `pnpm install --frozen-lockfile` — ok
- `pnpm run typecheck` — pass
- `pnpm run build` — pass
- `npx vitest run` — 22 files / 259 tests passed (includes 11 new discriminated-union cases)

(No lint script exists in open-live.)


birme (commented):
**pr-reviewer verdict: APPROVE** (recorded as a comment — GitHub blocks a formal `--approve` because reviewer and author share this automation's single `birme` token identity; branch protection here requires only the green CI check, per the deliberate decision in agent-team-open-live#4).

Discriminated-union variants match the `MACRO_EXEC` executor fields exactly (`src/ws/controller.ts:1023-1084`); `.strict()` correctly turns cross-type field mixing into a 400 (verified zod semantics — without it unknown keys are silently stripped); no previously-valid single-type payload is newly rejected; tests cover all 6 valid variants plus 5 rejection cases; CI `build-and-test` green.

PR #259, opened 2026-09-15
fix(db): add findTrusted opt-out so literal operator selectors bypass the Mango guard

### LINKED ISSUE #257: Mango-injection guard blocks the application's own queries — activation fails with any output assigned
# Mango-injection guard blocks the application's own queries — activation fails with any output assigned

**Repo:** Eyevinn/open-live · **Type:** Bug · **Severity:** High — main is broken for any production with an output

## Problem

Activating a production that has an output assigned fails with HTTP 500:

```
MangoInjectionError: Unsafe Mango operator '$in' in find() selector
  at safeSelector (dist/db/index.js:40)
  at Proxy.<anonymous> (dist/db/index.js:71)
  at productions.js:532
POST /api/v1/productions/{id}/activate → 500
```

Reproduced on two separate OSC instances and locally against latest `main`. The trigger is solely the presence of an output assignment — the port, the mode (listener/caller) and the source set make no difference.

| Output assigned | Result |
|---|---|
| none | activates cleanly |
| any | 500 |

## Impact

Three code paths are affected, all of which use Mango operators in selectors written by the application:

| Site | Selector | Effect today |
|---|---|---|
| `src/routes/productions.ts:607` | `status: { $in: ['active','activating'] }` | **Activation 500s** whenever `outputAssignments` is non-empty |
| `src/routes/sources.ts:213` | `$in` + `$elemMatch` | Deleting a source fails its in-use check |
| `src/services/idle-watchdog.ts:67` | `status: { $in: [...] }` | Watchdog cannot seed active productions at startup — logs `Failed to seed active productions`, then runs blind to anything already active |

## Root cause

`#237` wired `safeSelector()` onto the `find()` hot path via the proxy in `withTypeGuard` (`src/db/index.ts`). The guard throws on **any** `$`-prefixed key, with no way to distinguish a selector written as a literal in application code from one built out of request data.

The guard's own doc comment states the assumption that made this look safe:

> All current callers pass hardcoded, operator-free selectors (e.g. `{ type: 'production' }`), so it is a no-op for them

That was already untrue when written — the three call sites above use operators.

Injection risk comes from a selector's **provenance**, not from whether it contains operators. `{ status: { $in: ['active','activating'] } }` written as a literal is harmless; `{ name: req.query.name }` is dangerous even with no `$` in the source, because the value may itself carry operators. The guard should keep catching the second and stop blocking the first.

## Proposed fix

Keep `find()` exactly as strict as it is today. Add one explicit, greppable opt-out for selectors written by application code.

In the `withTypeGuard` proxy in `src/db/index.ts`, alongside the existing `find` branch:

```ts
if (prop === 'findTrusted') {
  // Selectors written as literals in application code, which may use Mango
  // operators. NEVER call this with a selector that contains, or is built
  // from, request data — use find() for anything user-derived.
  return (query: Nano.MangoQuery, ...rest: unknown[]) => boundFind(query, ...rest);
}
```

The safety property is that the exemption is **named at the call site**: `grep -rn findTrusted src/` yields the complete audit list of queries that bypass the guard — three lines. Any future endpoint that forwards user input uses plain `find()` and stays protected, which was the point of `#64`.

### Rejected alternative

Allow-listing specific operators (`$in`, `$elemMatch`) while rejecting `$regex`/`$where` is weaker: an attacker who can smuggle `$in` can still widen a query past its intended predicate, and the list needs extending each time application code reaches for a new operator.

## Files affected

| File | Change |
|---|---|
| `src/db/index.ts` | Add the `findTrusted` branch; add `GuardedScope<T>` interface extending `Nano.DocumentScope<T>`; update the return type of `withTypeGuard` and the five accessors (`getDb`, `getSourcesDb`, `getConfigsDb`, `getGraphicsDb`, `getOutputsDb`); correct the misleading doc comment |
| `src/routes/productions.ts:607` | `find` → `findTrusted` |
| `src/routes/sources.ts:213` | `find` → `findTrusted` |
| `src/services/idle-watchdog.ts:67` | `find` → `findTrusted` |
| `src/__tests__/safe-selector.test.ts` | Add coverage for the trusted path |

Roughly 15 lines of behavioural change. No API change, no change to any user-facing endpoint.

## Acceptance criteria

- [ ] Activating a production with an output assigned succeeds
- [ ] Deleting a source runs its in-use check without throwing
- [ ] `[idle-watchdog] Failed to seed active productions` no longer appears at startup
- [ ] Plain `find()` still rejects `$in`, `$or`, `$regex`, and operators nested in arrays or smuggled into values — existing tests unchanged and passing
- [ ] `grep -rn findTrusted src/` returns exactly the three intended call sites

## Test plan

1. **Unit** — extend `safe-selector.test.ts`: `findTrusted` passes an operator selector through; `find` still throws. Existing assertions untouched.
2. **Manual, locally** — against the shared OSC CouchDB and shared Strom: assign an output to a production and activate. Currently 500s; must reach `Production activated — flow playing`.
3. **Regression gap to note** — every existing test mocks the database, which is why this shipped green. A CI integration layer with a real CouchDB would have caught it on the PR. Out of scope here; proposed separately.

## Out of scope

- The CI integration-test layer (separate ticket)
- Revisiting whether `safeSelector` is the right shape of guard at all — this keeps `#237`'s design and closes the gap in it

## Notes

`#237` is recent work by @birme. This is not a revert: the guard stays, and the fix is the escape hatch it needs to coexist with the application's own queries. Worth his review given he wrote the original.

birme:
## Summary
- Activation of any production with an output assigned currently 500s with `MangoInjectionError` because the #237 `safeSelector()` guard rejects every `$`-prefixed selector key, including literal operators written by the application itself.
- Adds a named, greppable `findTrusted` opt-out on the `withTypeGuard` proxy that skips the guard for selectors written as literals in application code, while `find()` stays exactly as strict as before (preserving the #64 injection backstop).

Closes #257

## Changes
- `src/db/index.ts`: add a `findTrusted` branch to the `withTypeGuard` proxy that calls the bound `find` without `safeSelector`; add `GuardedScope<T>` interface extending `Nano.DocumentScope<T>`; update the return types of `withTypeGuard` and the five accessors (`getDb`, `getSourcesDb`, `getConfigsDb`, `getGraphicsDb`, `getOutputsDb`); export `withTypeGuard` for testing; correct the misleading doc comment that claimed all callers pass operator-free selectors.
- `src/routes/productions.ts`: migrate the `status: { $in: [...] }` activation conflict query to `findTrusted`.
- `src/routes/sources.ts`: migrate both literal `$in`/`$elemMatch` in-use checks (active and inactive-production paths) to `findTrusted`.
- `src/services/idle-watchdog.ts`: migrate the `status: { $in: [...] }` seed query to `findTrusted`.
- `src/__tests__/safe-selector.test.ts`: add coverage that `findTrusted` passes operator selectors through and that plain `find` still throws on `$in`/`$or`/`$regex`/`$elemMatch`; existing `safeSelector` assertions untouched.

Provenance, not presence of operators, drives injection risk — every bypass is now explicit and greppable (`grep -rn findTrusted src/` yields exactly the four call sites plus the definition).

## Test Plan
- [x] Unit tests pass (`npx vitest run` — 266 passed)
- [x] `pnpm run typecheck` passes
- [x] `pnpm run build` passes
- [ ] Manual verification: assign an output to a production and activate against a real CouchDB (reaches `Production activated`, no 500); delete an in-use source (409 without throwing); watchdog seeds active productions at startup with no `Failed to seed active productions` log

## Checklist
- [x] Code-quality checks passed (install/typecheck/build/tests green)
- [x] No hardcoded SRT ports
- [x] GStreamer pipelines have watchdogs (if applicable) — n/a
- [x] EFP sequence validation present (if applicable) — n/a
- [x] `find()` remains strict; only the named `findTrusted` opt-out bypasses the guard

🤖 Generated with Claude Code


birme:
## pr-reviewer verdict: APPROVE

Reviewed by `pr-reviewer` (a different agent than the implementer). Verified against the checked-out branch, not just the description.

1. **`find()` backstop intact** — `src/db/index.ts` `find` branch still calls `safeSelector(query.selector)` before forwarding; `findTrusted` is the *only* bypass and is a distinct, greppable named method. The #64 injection guard is preserved.
2. **Exactly the four intended literal-operator call sites migrated**, none forwarding user data into selector *keys*: `productions.ts:606-607` (`$in`), `sources.ts:212-213` (`$in`+`$elemMatch`), `sources.ts:223-224` (`$elemMatch`), `idle-watchdog.ts:66-67` (`$in`). The two `sources.ts` selectors embed `req.params.id` only in *value* position under an equality match — not an injection vector.
3. `grep -rn findTrusted src/` returns only the definition + 4 call sites + tests. No stray usages.
4. Tests strengthened, not weakened: original `safeSelector` unit tests untouched; new proxy-level coverage confirms `find()` still throws on `$in`/`$or`/`$regex`/nested `$elemMatch` while `findTrusted()` passes operators through.
5. Pre-PR checks all pass: typecheck clean, build clean, `npx vitest run` = 266/266.
6. Real CI `build-and-test`: pass.

Non-blocking: the "never call with request-derived data" contract is enforced by convention + greppability, not the type system — fine here since every bypass is named at the call site.

birme:
🤖 Housekeeping (daily-backlog-pr Phase 3.5): the target issue **#257 is already closed** by **#262** ("fix(db): let application selectors use Mango operators", merged 2026-09-15), which lands a competing fix for the same `MangoInjectionError`-on-activation bug. This branch is now `CONFLICTING` against `main` for that reason. Reviewed both approaches: #262 is the merged winner, so this PR is a stale duplicate. Closing. If you believe the `findTrusted`-opt-out design here is materially safer than #262's approach, reopen with a rebase and a note on the concrete difference.

